PR #79, opened 2023-03-29
Add option to start synchronization from prepared db

pawellewandowski98:
Description:
---
- Added prepared, synchronized and compressed db
- Added viper env which allows to use prepared db

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -43,8 +43,6 @@ func main() {
 	vconfig := vconfig.NewViperConfig(appname).
 		WithDb()
 
-	freshDbIfConfigured()
-
 	// Unzip prepared db file if configured.
 	if viper.GetBool(preparedDb) {
 		err := os.Remove(viper.GetString(dbFileEnv))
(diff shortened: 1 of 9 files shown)


cmd/main.go
```diff
@@ -34,13 +36,27 @@ import (
...
+const preparedDbFilePath = "db.preparedDbFile.path"
 
 func main() {
 	vconfig := vconfig.NewViperConfig(appname).
 		WithDb()
 
 	freshDbIfConfigured()
 
+	// Unzip prepared db file if configured.
+	if viper.GetBool(preparedDb) {
```
proapi, line 49: Please update the readme to show that there is a possibility to run it with an additional flag and it will synchronize

sirdeggen (dismissed):
Works perfectly, thanks.

As @proapi stated in the comments, please add the new functionality to the readme making clear that by default it will assume you already have a db in place.

Once more request - can you change the docker-compose.yaml - I'll make a comment on the specific line.

cmd/main.go
pawellewandowski98: Thanks! Added two new variables to readme

proapi (approved):

PR #80, opened 2023-03-29
Update docker-compose.yml

sirdeggen:
Ensure the default mount is in place because permissions may not allow mounting at the home directory.


proapi:
Closing, we will update the previous PR with proper code, not mounting it with the shared volume, since it is not needed.

PR #88, opened 2023-04-03
Fix http endpoints

pawellewandowski98:
Description:
---
- Add new structs BlockHeaderResponse and BlockHeaderStateResponse to transport/http package
- Changed endpoints in http server, now they return data in new structs
- Move DbBlockHeader struct and methods to repository package

### DIFF
--- a/repository/domains/headers.go
+++ b/repository/domains/headers.go
@@ -1,0 +1,95 @@
+package domains
+
+import (
+	"math/big"
+	"strconv"
+	"time"
+
+	"github.com/libsv/bitcoin-hc/domains"
+	"github.com/libsv/bitcoin-hc/internal/chaincfg/chainhash"
+)
+
+// DbBlockHeader represent header saved in db.
+type DbBlockHeader struct {
+	Height           int32     `db:"height"`
+	Hash             string    `db:"hash"`
+	Version          int32     `db:"version"`
+	MerkleRoot       string    `db:"merkleroot"`
+	Timestamp        time.Time `db:"timestamp"`
+	Bits             uint32    `db:"bits"`
+	Nonce            uint32    `db:"nonce"`
+	State            string    `db:"header_state"`
+	Chainwork        string    `db:"chainwork"`
+	CumulatedWork    string    `db:"cumulatedWork"`
+	PreviousBlock    string    `db:"previousblock"`
+	DifficultyTarget uint32    `db:"difficultytarget"`
+}
+
+// ToBlockHeader converts work from string to big.Int and return BlockHeader.
+func (dbh *DbBlockHeader) ToBlockHeader() *domains.BlockHeader {
+	if dbh.CumulatedWork == "" {
+		dbh.CumulatedWork = "0"
+	}
+	cumulatedWork, ok := new(big.Int).SetString(dbh.CumulatedWork, 10)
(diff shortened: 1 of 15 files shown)


transports/http/server/http_server.go
```diff
@@ -20,8 +20,8 @@ func NewHttpServer(port int, handler http.Handler) *HttpServer {
 		httpServer: &http.Server{
 			Addr:         ":" + fmt.Sprint(port),
 			Handler:      handler,
-			ReadTimeout:  5 * time.Second,
-			WriteTimeout: 5 * time.Second,
+			ReadTimeout:  10 * time.Second,
```
wregulski, line 23: Maybe consider putting this value "10" in config?

transports/http/handlers/headers_handler.go
```diff
@@ -57,63 +61,76 @@ func (h *Handler) getHeaderByHeight(c *gin.Context) {
...
 func (h *Handler) getHeadersState(c *gin.Context) {
 	hash := c.Param("hash")
-	state, err := h.services.Headers.GetHeadersState(hash)
+	header, err := h.services.Headers.GetHeaderByHash(hash)
 
 	if err == nil {
-		c.JSON(http.StatusOK, state)
+		confirmations := h.services.Headers.CalculateConfirmations(header)
+		// h.services.Headers.CalculateConfirmations(header)
+		// fmt.Println("confirmations")
```
wregulski, line 125: Please remove comments

wregulski (changes requested):

transports/http/domains/headers.go
```diff
@@ -0,0 +1,65 @@
+package domains
```
dorzepowski, line 1: IMHO there is no need to put it in a separate package, it could be part of the http package.

repository/domains/headers.go
```diff
@@ -0,0 +1,95 @@
+package domains
```
dorzepowski, line 1: IMHO there is no need to put it in a separate package, it could be part of the repository package.

pawellewandowski98: Can not move this to repository package because of import cycle with `/data/sql/headers.go`

dorzepowski: So maybe just rename it to model or something that doesn't mislead that it's part of the domain logic?

transports/http/server/http_server.go
pawellewandowski98: Thanks! I added new envs in viper

transports/http/handlers/headers_handler.go
pawellewandowski98: Done

transports/http/domains/headers.go
pawellewandowski98: Done

repository/domains/headers.go
pawellewandowski98: Ok I renamed it to `dto`

wregulski (dismissed):

dorzepowski (approved):

sirdeggen (approved):
This should go in, definitely an improvement. There are small fixes still to make. I'll raise issues describing them clearly because they are subtle.

sirdeggen:
rebased on master, good to merge

PR #100, opened 2023-04-17
map state response

sirdeggen:
Tips ought to be formatted including state, chainWork, and height.


pawellewandowski98:
Changes for this issue are now in this PR: https://github.com/libsv/pulse/pull/104

PR #106, opened 2023-04-27
Add authorization for http server

### LINKED ISSUE #105: [FEATURE] Authentication for endpoints
**Is your feature request related to a problem? Please describe.**
Right now we allow public access to endpoints in PULSE. If we would like to setup this service somewhere on the server, everyone would be able to make requests, which from a security point of view shouldn't be allowed.

**Describe the solution you'd like**
There should be two ways to authenticate in PULSE:
1. 
A good solution would be a simple Bearer token.
The app should have an ENV added which is “token” to connect to it. That’s the first bearer token. The app checks every request if it contains the `Authorization: Bearer "token"` and either 401 or responses as before.

2. 
We should also add an endpoint which adds tokens to a live instance.
`POST /api/access`
Which adds a random access token to a database table and memory.

There should also be a possibility to revoke the access token, so another endpoint is needed:
`DELETE /api/access`

Both endpoints to create and delete access tokens should be already secured by the bearer token from the first point.
Checking which token should be done first for the one in the env variable then checking others in the database. The one in env is more important.

pawellewandowski98:
Related links:
---

Close #105 

Description:
---
- Added authorization for every endpoint in http server
- Added authorization to swagger
- Added two endpoints:
-     [GET]`/access` Used to generate new access token
-     [DELETE] `/access/<token>` Used to revoke access token
This two endpoints are accessible only with env authorization token


wregulski (approved):

docker-compose.yml
sirdeggen, line 1: Here I added a default admin token for people to see at the config level.

sirdeggen (dismissed):

sirdeggen (approved):

Issue #107, opened 2023-05-01
[FEATURE] Tips Endpoint Speed Improvement


sirdeggen:
**Tips Endpoint Speed Improvement**
Sometimes the response will timeout because calculating the response for the tips endpoint takes a long time. The SQL query is unnecessarily scanning the whole table to get what is almost always one record. 

**Describe the solution you'd like**
Tips will only be considered at all if they are within the top few headers anyway, realistically, so perhaps the practical number to check would be the 10 most recently added headers, or something to this effect.

**Describe alternatives you've considered**
There could be issues with this in that one could receive a bunch of headers from an attacker which could mean your tips could become out of whack with reality somehow. In all normal circumstances this would not be the case though.

**Additional context**
In production I've seen 9 second response time which is just lame. It's likely to increase with height.

PR #108, opened 2023-05-05
Much more performant tips endpoint

sirdeggen:
Enhances [performance issue](https://github.com/libsv/pulse/issues/107).

Achieved by limiting the query to the top 100 headers, and performing the "tips only" logic on that small set.

100 was chosen because any tips prior to 100 blocks back in history should be rejected anyway as this is most likely an attack.

Response time is down from 8 full seconds to 340ms.

### DIFF
--- a/data/sql/headers.go
+++ b/data/sql/headers.go
@@ -120,6 +120,13 @@ const (
     WHERE height = ?
     `
 
+	sqlSelectHighestHundred = `
+	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
+	FROM headers
+	ORDER BY height DESC
+	LIMIT 100
+	`
+
 	sqlSelectTips = `
 	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, cumulatedWork, header_state
 	FROM headers
@@ -164,6 +171,16 @@ func NewHeadersDb(db *sqlx.DB, dbType vconfig.DbType) *HeadersDb {
 			},
 		},
 	}
+}
+
+func contains(slice []string, item string) bool {
+	set := make(map[string]struct{}, len(slice))
+	for _, s := range slice {
+		set[s] = struct{}{}
+	}
+
+	_, ok := set[item]
+	return ok
 }
 
 // Create method will add new record into db.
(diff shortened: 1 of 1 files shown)


data/sql/headers.go
```diff
@@ -323,13 +333,25 @@ func (h *HeadersDb) GetAncestorOnHeight(hash string, height int32) (*dto.DbBlock
 	return bh[0], nil
 }
 
-// GetAllTips returns all tips from db.
+// GetAllTips returns headers whose hash does not appear as previous block hashes in any header within the highest 100 headers.
```
dorzepowski, line 336: @sirdeggen what actually is a business requirement for that method 
What we want to actually return from endpoint in an imaginary ideal world?
Which of the following should be included in the ideal solution:
1. tip of longest chain
2. tip of newest forked chain
3. tip of all forked chains
4. tip of all orphan chains

data/sql/headers.go
```diff
@@ -120,12 +120,12 @@ const (
...
-	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, cumulatedWork, header_state
+	sqlSelectHighestHundred = `
+	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
 	FROM headers
-	WHERE hash NOT IN (SELECT previousblock
-					   FROM headers)
-				   `
+	ORDER BY height DESC
+	LIMIT 100
+	`
```
dorzepowski, line 128: One think still bothers me. So i went after my idea, and I think that changing only this SQL to such a query should do the work and we will still get all tips of all concurrent chains and orphans:

```suggestion
sqlSelectTips = `
with mainTip as (
select hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
from headers
where header_state = 'LONGEST_CHAIN'
order by height
limit 1
)
select hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
from mainTip
union
select hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
from headers
where header_state != 'LONGEST_CHAIN' and
        hash not in (select previousblock from headers where header_state != 'LONGEST_CHAIN')
        `
```

dorzepowski: For the sake of completeness, I prepared a separate PR with this proposition:
https://github.com/libsv/pulse/pull/112
Please check if this fits your needs @sirdeggen 

And sorry if this looks a little rude that I'm mingling yours PR.

sirdeggen: Thanks, will take a look now

sirdeggen: Good, thanks.

sirdeggen:
Fixed in alternative branch thanks to @dorzepowski

Issue #109, opened 2023-05-05
[BUG] Difficulty Target Never Stored


sirdeggen:
**Describe the bug**
difficultytarget is never stored in the database because none of the SQL commands reference it. Therefore any time it is in a response it is zero.

This needs to be added, and the blockheaders.xz needs to be dumped and recreated from scratch to include this info.

PR #112, opened 2023-05-08
perf(#109): increase performence listing all tips

dorzepowski:


data/sql/headers.go
```diff
@@ -121,10 +121,20 @@ const (
...
 	sqlSelectTips = `
-	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, cumulatedWork, header_state
-	FROM headers
-	WHERE hash NOT IN (SELECT previousblock
-					   FROM headers)
+	with mainTip as (
+	select hash, height, version, merkleroot, nonce, bits, chainwork, previousblock, timestamp, header_state, cumulatedWork
+	from headers
+	where header_state = 'LONGEST_CHAIN'
+	order by height
```
sirdeggen, line 128: It doesn't work, so I assume what's happening is this should be DESC order - I'll try

sirdeggen: Confirmed - approved.

sirdeggen (approved):
Required inversion of height order but otherwise a tidy solution thank you.

PR #113, opened 2023-05-10
tests: common method for setup e2e tests

dorzepowski:
Additionally to create a new way of testing endpoint I also make some changes in access/ endpoints (because I misunderstood how it works and I start to fix the code to match tests requirements)

Changes made:
1. Testing:
   a. provide utility to start and eventually configure pulse for automatic tests
2. access tokens
  a. change http method for generating tokens from GET to POST
  b. provide method to check token with GET /access
  c. move middleware to separated package and limit its responsibility to only extract a token from header (and check if it's valid) and put a token in gin context
  d. provide a handler wrapper which can be used to limit access to an endpoint only for admins (the one who send admin token in Authorization Bearer). Example use is on a POST /access - endpoint to create additional tokens.


pawellewandowski98 (approved):

wregulski (dismissed):

README.md
```diff
@@ -129,11 +129,64 @@ or with Docker
...
 
+### Authentication
+
+#### Enabling
+
+To setup and enable authentication with token setup the following environment variables
+- HTTP_SERVER_USEAUTH=true
+- HTTP_SERVER_AUTHTOKEN=replace_me_with_token_you_want_to_use_as_admin_token
+
+ℹ️ actually authorization is enabled by defautl so only the HTTP_SERVER_AUTHTOKEN env is required
```
pawellewandowski98, line 148: typo in default

README.md
```diff
@@ -129,11 +129,64 @@ or with Docker
...
 
+### Authentication
+
+#### Enabling
+
+To setup and enable authentication with token setup the following environment variables
+- HTTP_SERVER_USEAUTH=true
+- HTTP_SERVER_AUTHTOKEN=replace_me_with_token_you_want_to_use_as_admin_token
+
+ℹ️ actually authorization is enabled by default so only the HTTP_SERVER_AUTHTOKEN env is required
```
sirdeggen, line 148: I would update this in line with what is said in it - seems like this was written in hindsight.

Above do not include `HTTP_SERVER_USEAUTH=true` there's then no need to mention it. The HTTP_SERVER_USEAUTH=false below is sufficient to explain.

sirdeggen (approved):
Looks great, thanks

PR #114, opened 2023-05-11
Add webhooks

### LINKED ISSUE #110: [FEATURE] Webhook solution to get informed about newest header received by pulse
**Is your feature request related to a problem? Please describe.**
When integrating with pulse I want to be informed that it received a new header.

**Describe the solution you'd like**
There should be exposed 2 endpoints:
1. to register a webhook by providing URL and eventually some token for authentication solution
2. to unregister a webhook

When a webhook is registered, everytime a new header is stored by Pulse, it should make a request on a registered URL with data of the newest header.

pawellewandowski98:
Related links:
---

Close #110 

Description:
---
- Added webhooks
- Added two endpoints:
-     [POST]`/webhook` Used to register or refresh new webhook
-     [DELETE] `/webhook?url=<url>` Used to revoke exisiting web hook

When new header appear, every registered and active webhook will be informed about it


transports/http/handlers/webhooks_handler.go
```diff
@@ -0,0 +1,93 @@
...
+		var tHeader, token string
+
+		// If custom header is specified, use it, otherwise use default
+		if strings.ToLower(reqBody.RequiredAuth.Type) == "custom_header" {
+			tHeader = reqBody.RequiredAuth.Header
+			token = reqBody.RequiredAuth.Token
+		} else {
+			tHeader = "Authorization"
+			token = "Bearer " + reqBody.RequiredAuth.Token
+		}
```
dorzepowski, line 45: IMHO that is a part of the logic of creating a webhook, so it rather should be placed in a service.
I think also that validation could be part of the create method in service.

transports/http/handlers/webhooks_handler.go
```diff
@@ -0,0 +1,93 @@
...
+	} else {
+		c.JSON(http.StatusBadRequest, err.Error())
+	}
+}
+
+func (h *Handler) initRegisteredWehooksRoutes(router *gin.RouterGroup) {
+	webhooks := router.Group("")
+	{
+		webhooks.POST("/webhook", h.registerWebhook)
+		webhooks.DELETE("/webhook", h.revokeWebhook)
```
dorzepowski, line 91: IMHO it's good idea to have an endpoint to list a webhooks so you can check the status of it.

service/webhooks_service.go
```diff
@@ -0,0 +1,100 @@
...
+// WebhookMaxTries is the maximum number of times a webhook will be retried.
+const WebhookMaxTries = "webhook.maxTries"
+
+// WebhooksService represents Webhooks service and provide access to repositories.
+type WebhooksService struct {
+	repo *repository.Repositories
+}
+
+// GenerateWebhook generates and save new webhook.
+func (s *WebhooksService) GenerateWebhook(url, tHeader, token string) (*domains.Webhook, error) {
```
dorzepowski, line 23: IMHO `RegisterWebhook` or `CreateWebhook` would be better name.

service/webhooks_service.go
```diff
@@ -0,0 +1,100 @@
...
+	webhook := domains.CreateWebhook(url, tHeader, token)
+	err := s.repo.Webhooks.AddWebhookToDatabase(webhook)
+	if err != nil {
+		w, _ := s.repo.Webhooks.GetWebhookByUrl(url)
```
dorzepowski, line 27: if there will be any database issue we will return invalid error info that 
`"webhook already exists and is active"`.

IMHO we should handle the error here first

service/webhooks_service.go
```diff
@@ -0,0 +1,100 @@
...
+		if w != nil && !w.Active {
+			err = s.repo.Webhooks.UpdateWebhook(w, w.LastEmitTimestamp, w.LastEmitStatus, 0, true)
+			w.Active = true
+			w.ErrorsCount = 0
+			return w, err
```
dorzepowski, line 32: Couldn't it be done like that:
```suggestion
			w.Active = true
			w.ErrorsCount = 0
			err = s.repo.Webhooks.UpdateWebhook(w)
			return w, err
```

service/webhooks_service.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		return nil, errors.New("webhook already exists and is active")
```
dorzepowski, line 34: Extract a method from that for example
```suggestion
		s.refreshWebhook(webhook)
```

service/webhooks_service.go
```diff
@@ -0,0 +1,100 @@
...
+				err = s.repo.Webhooks.UpdateWebhook(webhook, timestamp, lastEmitStatus, errorsCount, active)
+			} else {
+				// If status code is 200, reset errors count and set active to true
+				err = s.repo.Webhooks.UpdateWebhook(webhook, timestamp, lastEmitStatus, 0, true)
+			}
+
+			if err != nil {
+				return err
+			}
+		}
```
dorzepowski, line 92: I would rather put most of that logic into a domain.Webhook, 
to have here something like:
```suggestion
		maxFailures := viper.GetInt(WebhookMaxTries) //actually this could probably be set at the service creation
		err := webhook.Notify(h, maxFailures)
		if err != nil {
		  s.log.Errorf(/* put some debugging info here*/)
		}
		s.repo.Webhooks.UpdateWebhook(webhook)
```

domains/webhooks.go
```diff
@@ -0,0 +1,66 @@
...
+		return 0, "", err
+	}
+
+	// Add the necessary headers.
+	req.Header.Add(w.TokenHeader, w.Token)
+	req.Header.Add("Content-Type", "application/json")
+
+	// Send the request.
+	client := &http.Client{}
+	res, err := client.Do(req)
```
dorzepowski, line 43: I have a doubt about using http in domain.
AFAIU in hexagonal arch we should probably pass here some component (implementation of the interface) that will handle the http request.
For example something like:
```golang
type WebhookTarget interface {
  Call(headers map[string][string], method http.MethodPost, url w.Url, body)
}
```

and implement it like here
and then pass such a component to Notify method.
So something like: 
`Notify(h *BlockHeader, target WebhookTarget)`

sirdeggen: In which case, I'm not sure WebhookTarget is the correct name. Webhook implies http so perhaps "WebTarget" is the interface name. Just a suggestion.

sirdeggen (changes requested):
Really nice, agree with the sentiments detailed by @dorzepowski to update one or two names, move some logic from handler to service and pass in a WebTarget Interface - that last request can be left to another PR where sockets are introduced because I'm sure that task will raise some considerations.

domains/webhooks.go
dorzepowski: @sirdeggen I was thinking about some interface to hide webhooks and websockets under it,
but I need to wait until this PR will be merged so it will be clear for me what and where should be "interfaced".

domains/webhooks.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return 0, "", err
+	}
+
+	defer res.Body.Close() // nolint: all
```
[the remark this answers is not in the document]
pawellewandowski98: As usual when methods `Close` are used in defer function, error is not checked

domains/webhooks.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		// Update the webhook after failed notification.
+		w.updateWebhookAfterNotification(0, "", err)
+		return err
+	}
+
+	defer res.Body.Close() // nolint: all
+
+	// Read the response.
+	body, _ := io.ReadAll(res.Body)
```
dorzepowski, line 51: you should assign err here 
```suggestion
	body, err := io.ReadAll(res.Body)
```

transports/http/webhooks_handler_test.go
```diff
@@ -0,0 +1,111 @@
...
+	RequiredAuth: handler.RequiredAuth{
+		Type:  "BEARER",
+		Token: "test-token",
+	},
+}
+
+// TestCreateWebhookEndpoint tests the webhook registration.
+func TestCreateWebhookEndpoint(t *testing.T) {
+	//setup
+	pulse, cleanup := testpulse.NewTestPulse(t, testpulse.WithoutApiAuthorization())
```
dorzepowski, line 29: IMHO in tests, authorisation could be disabled by default.
And probably it is.
So this could be just:

```suggestion
	pulse, cleanup := testpulse.NewTestPulse(t)
```

service/webhooks_service.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+
+	if w != nil && !w.Active {
+		w.Active = true
+		w.ErrorsCount = 0
+		err = s.repo.Webhooks.UpdateWebhook(w)
+		return w, err
+	}
+	return nil, errors.New("webhook already exists and is active")
```
dorzepowski, line 98: I don't think that it's good idea to return an error here.
If we succeed to update a webhook then it is success and we should just return an update webhook.

transports/http/handlers/webhooks_handler.go
```diff
@@ -0,0 +1,105 @@
...
+		if err == nil {
+			c.JSON(http.StatusOK, webhook)
+		} else if webhook == nil {
+			c.JSON(http.StatusOK, err.Error())
+		}
+	}
+
+	if err != nil {
+		c.JSON(http.StatusBadRequest, err.Error())
+	}
```
dorzepowski, line 41: IMHO reversing the error checks (I mean first checking if there is an error)
is more Go common and actually looks more clear.

```suggestion
	var reqBody WebhookRequest
	if err := c.Bind(&reqBody); err != nil {
		c.JSON(http.StatusBadRequest, err.Error())
		return
	}


	if reqBody.Url == "" {
		c.JSON(http.StatusBadRequest, "Url is required")
		return
	}

	webhook, err := h.services.Webhooks.CreateWebhook(reqBody.RequiredAuth.Type, reqBody.RequiredAuth.Header, reqBody.RequiredAuth.Token, reqBody.Url)
	if err != nil {
		c.JSON(http.StatusBadRequest, err.Error())
		return
	} 
	
	c.JSON(http.StatusOK, webhook)
```

dorzepowski (approved):

sirdeggen (approved):
LGTM

PR #115, opened 2023-05-16
fix: zero difficulty target

vitalibalashka:
Zero difficulty target issue, go fmt and small code formatting


vitalibalashka:
@sirdeggen, please review.
I've found that difficulty target actually was always stored, but not as difficultytarget, but as bits. And now difficultytarget in response will contain this value from bits
Screenshot 2023-05-16 at 13 34 16

sirdeggen (approved):
Well found!

PR #134, opened 2023-09-27
feat(BUX-185): p2pConfig refactoring

witalij-4chain:
P2PConfig refactoring.
From now on, the viper is responsible for the p2p config.
p2p ini config replaced with yaml. Each param in p2pconfig can be rewritten by env variable.
The ability to override p2p config through flags with the highest priority is left. Example: if proxy is specified via flag, via yaml and via env variable, flag value will be used.
Global p2p config is deleted, each service which require this config now has pointer to it or pointer to some config value.


vconfig/p2pconfig/p2p_cli_config.go
```diff
@@ -0,0 +1,106 @@
...
+		_, err := parser.AddGroup("Service Options", "Service Options", so)
+		if err != nil {
+			cfg.Logger.Error(err)
+		}
+	}
+	return parser
+}
+
+// ParseFlags returns Config from flags.
+func ParseFlags(customWorkingDirectory string) (*Config, error) {
```
kuba-4chain, line 37: If I understand correctly, `customWorkingDirectory` means a directory where a default config file for p2p is. Maybe instead of putting it as function parameter here and in `DefaultP2PConfig()` function, we could use it as a normal flag and/or env variable just as other config options?

Because now, we have to put an empty string into the parameter for default values and it feels weird.

Also, the name doesn't suggest that it's a directory for storing config file, so maybe something like `configFileDirectory` or `customConfigDirectory` would be better? 

WDYT?

vconfig/p2pconfig/p2p_cli_config.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
+		}
+		os.Exit(0)
+	}
+	return &cfg, nil
+}
+
+// DefaultP2PConfig returns default p2pConfig.
+func DefaultP2PConfig(customWorkingDirectory string) Config {
```
kuba-4chain, line 85: Same as above, it would be really nice to get rid of this function parameter :)

kuba-4chain (commented):
Awesome job!

I just have one comment below :)

Also, after this is merged, we should update the documentation in our pulse wiki with the new config options, where they are and how to set them.

vitalibalashka:
> Awesome job!
> 
> I just have one comment below :)
> 
> Also, after this is merged, we should update the documentation in our pulse wiki with the new config options, where they are and how to set them.

The docs are in progress, I've just tired of this Pulse and I've seen the much more important issue with go-paymail. So I've switched to it, but after it I'll update the docs

vconfig/p2pconfig/p2p_cli_config.go
vitalibalashka: This solution with `customWorkingDirectory` is taken from BUX. It's necessary mostly for tests, not for regular usage. If you want to use a custom you can specify by --configfile or -C flag.
Regarding the name, you're right. Not exactly a working directory, it is more rootDirectory

data/sql/headers.go
```diff
@@ -182,6 +183,7 @@ func NewHeadersDb(db *sqlx.DB, dbType vconfig.DbType) *HeadersDb {
 				insertBH: sqliteInsertHeader,
 			},
 		},
+		logger: lf.NewLogger("headers-db"),
```
dorzepowski, line 186: can you name it simply `log`?
```suggestion
		log: lf.NewLogger("headers-db"),
```

transports/http/server/http_server.go
```diff
@@ -33,6 +35,7 @@ func NewHttpServer(port int) *HttpServer {
 			WriteTimeout: time.Duration(viper.GetInt(vconfig.EnvHttpServerWriteTimeout)) * time.Second,
 		},
 		handler: handler,
+		logger: lf.NewLogger("http"),
```
dorzepowski, line 38:
```suggestion
		log: lf.NewLogger("http"),
```

vconfig/viper.go
```diff
@@ -66,3 +82,31 @@ func setWebsocketDefaults() {
...
+	viper.SetDefault(EnvP2PBanDuration, p2pconfig.DefaultBanDuration)
+	viper.SetDefault(EnvP2PLogDir, p2pconfig.DefaultLogDir)
+	viper.SetDefault(EnvP2PExcessiveBlockSize, p2pconfig.DefaultExcessiveBlockSize)
+	viper.SetDefault(EnvP2PTrickleInterval, p2pconfig.DefaultTrickleInterval)
+	viper.SetDefault(EnvP2PBlocksForForkConfirmation, p2pconfig.DefaultBlocksToConfirmFork)
+}
+
+// ParseP2PConfig init p2p viper config based from specific config file.
+func ParseP2PConfig(cliCfg *p2pconfig.Config) *p2pconfig.Config {
+	viper.SetConfigFile(cliCfg.ConfigFile)
```
dorzepowski, line 99: Maybe we should only read config file if the path is provided?
Or eventually assume some config location that is overriding the defaults, and don't raise error if it is not found?

vconfig/viper.go
```diff
@@ -66,3 +82,31 @@ func setWebsocketDefaults() {
...
+	if err := viper.ReadInConfig(); err != nil {
+		err = fmt.Errorf("config can't read %s", err.Error())
+		log.Fatal(err)
```
dorzepowski, line 102: I hate this method as it is missleading. IMHO we should have line for logging the error, and second line to panic

vconfig/viper.go
```diff
@@ -66,3 +82,31 @@ func setWebsocketDefaults() {
...
+	}
+
+	p2pCfg := new(p2pconfig.Config)
+
+	if err := viper.Unmarshal(&p2pCfg); err != nil {
```
dorzepowski, line 107: Maybe we could actually unmarshal whole configuration with nested groups, and stop accessing values by keys?

vitalibalashka: Done. Now we unmarshall all configs, not only p2p

vconfig/viper.go
vitalibalashka: Process was completely changed. Now we have flag --ignoreconfig to ignore reading of json config file

transports/http/server/http_server.go
vitalibalashka: Done

data/sql/headers.go
vitalibalashka: Done

cmd/main.go
```diff
@@ -83,40 +90,33 @@ func main() {
...
-	}
-
-	logger.SetLevelFromString(lf, configs.Cfg.DebugLevel)
-	logger.SetLevelFromString(log, configs.Cfg.DebugLevel)
+	logger.SetLevelFromString(lf, c.P2PConfig.LogLevel)
+	logger.SetLevelFromString(log, c.P2PConfig.LogLevel)
 
 	// Do required one-time initialization on wire
-	wire.SetLimits(configs.Cfg.ExcessiveBlockSize)
+	wire.SetLimits(c.P2PConfig.ExcessiveBlockSize)
```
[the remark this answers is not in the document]
vitalibalashka: Go philosophy says to use short names. Usage of one-letter variables is a common.
But I've renamed "config" to "cfg" here, and "P2PConfig" was renamed to "P2P"

cmd/main.go
```diff
@@ -49,7 +49,14 @@ const appname = "headers"
...
+
+	flagCfg, err := p2pconfig.ParseFlags("")
+	if err != nil {
+		log.Criticalf("failed to parse config from flags: %v\n", err)
+		os.Exit(1)
+	}
+
+	c := vconfig.NewViperConfig(appname, flagCfg).
 		WithDb().
 		WithAuthorization()
```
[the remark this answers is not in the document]
vitalibalashka: Done

cmd/main.go
```diff
@@ -49,7 +49,14 @@ const appname = "headers"
...
 	log := lf.NewLogger("main")
-	c := vconfig.NewViperConfig(appname).
+
+	flagCfg, err := p2pconfig.ParseFlags("")
+	if err != nil {
+		log.Criticalf("failed to parse config from flags: %v\n", err)
+		os.Exit(1)
+	}
+
+	c := vconfig.NewViperConfig(appname, flagCfg).
```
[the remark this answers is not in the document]
vitalibalashka: vconfig package was renamed to config. NewViperConfig renamed to Load

internal/tests/testpulse/test_pulse.go
```diff
@@ -66,20 +68,19 @@ func (p *TestPulse) When() *When {
...
-	conf := vconfig.NewViperConfig("test-pulse")
-	lf := testlog.NewTestLoggerFactory()
-
 	//override arguments otherwise all flags provided to go test command will be parsed by LoadConfig
 	os.Args = []string{""}
 
-	// Load configuration and parse command line.  This function also
-	// initializes logging and configures it accordingly.
-	err := configs.LoadConfig()
+	os.Args = []string{"test-pulse", "--configfile", getConfigPath(t)}
```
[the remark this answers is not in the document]
vitalibalashka: Added flag --ignoreconfig for such purpose.
os.Args duplicate is fixed

vconfig/p2pconfig/p2p_cli_config.go
vitalibalashka: Mechanism was changed. customWorkingDirectory can be forgotten like a bad dream.

vconfig/p2pconfig/p2p_cli_config.go
vitalibalashka: No more customWorkingDirectory. And this function was also deleted to be honest. There is only similar function in fixtures

vconfig/viper.go
vitalibalashka: Could you please clarify? I guess I don't fully understand what you want to say.
Panic will contain this error

vconfig/p2pconfig/p2p_cli_config.go
kuba-4chain: Nice!

kuba-4chain (dismissed):
Really nice design choices! 😎 

A lot has changed tho, so now we definitely need to update the config documentation 🥲

Maybe we could create a JIRA task for it and do it later, because I can imagine you have enough of Pulse for some time 😄

cmd/main.go
```diff
@@ -49,7 +49,14 @@ const appname = "headers"
 func main() {
 	lf := logger.DefaultLoggerFactory()
 	log := lf.NewLogger("main")
-	c := vconfig.NewViperConfig(appname).
+
+	flagCfg, err := p2pconfig.ParseFlags("")
```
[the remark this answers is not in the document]
dorzepowski: What do you mean, by  
> requiring mapping for everything

I thought that viper is handling all the ways we choose for configuration with predefined priority for each of the ways.
So where is this "hell"?

vconfig/viper.go
dorzepowski: > Fatal is equivalent to Print() followed by a call to os.Exit(1).

From my perspective this is messing up with purpose of function. Maybe it's cool that it shortens logging and exiting in single line.
But actually if you read it, you can quickly jump over this line, with the thinking that we're only logging here, but it's not true. We're making some big thing, we're shutting down whole program. But it reads like just log a message with Fatal level.

cmd/main.go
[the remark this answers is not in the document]
vitalibalashka: If you work with config as a struct you need to map each field with it's value in a file config.
On this stage everything is great, even if you work with nested json file. Just specify mapstructure tag.
Next stage: you want to override some param by env variable. Also ok. You just add a new replacer and you can provide values combining your mapstructure tags with underscore, for example "HTTP_PORT" or "P2P_PROXY" and so on.
But if you want to override something by flags you probably want to have them short, without underscores and so on, you probably want to provide them just like "--port" or "--proxy" or maybe by some short flag like "-P".
And here you need to register all your flags and bind them to viper. And it'll be a lot of dumb code.

vconfig/viper.go
vitalibalashka: fatal replaced with Print() and os.Exit(1)

cmd/main.go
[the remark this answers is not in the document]
dorzepowski: IMHO, the short flags are only needed for the most commonly used params, the rest can be in form "--p2p.something.something.else" or with `_` instead of `.`
I think that this is not a problem, and for me it seems like a way to go.

cmd/main.go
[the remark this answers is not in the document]
dorzepowski: 1. `cfg` name is ok.
2. I think that actually "Go philosophy says to use short names" is over-interpretation. I didn't saw a good practice in Go that was saying that you should use short names. The only think that the one-letter variable name rule is applied for is the name of the "receiver" of method.

cmd/main.go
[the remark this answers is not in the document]
dorzepowski: Or maybe I don't understand it correctly.
I assume that by default each property can be set via flags when using some default pattern to build those flags.
Or you need to register each and every flag?

dorzepowski (approved):

dorzepowski (changes requested):
IMHO the config file must be optional, otherwise, the pulse docker image and archives in release will stop working, and making it required, would force us to include the default json config in each of this artifacts.

dorzepowski (approved):

kuba-4chain (approved):
Everything tested, headers are syncing, endpoints are available, config works really well.

PR #153, opened 2023-11-09
feat(BUX-188): change database pre-loading

mwilkosinski:
# Pull Request Checklist

---

### What has been done in this PR

1. Refactorings:
   - Moved the CLI commands to a separate package (`cli/cli.go`). Previously they were handled directly in `config/viper.go`, which is not a good place for such functionality.
   - Moved all the code related to database management to the `database` folder. Created a `database` package containing the code for connecting to the database and running migrations. Also moved the existing `sql` package from `data/sql` to database, but without any changes.
   - Removed the database handling code from `main.go`
   - Converted the implementation for SQLite into an adapter, to enable support for other databases.
 2. New database preloading method:
     - The database is now being preloaded with headers data imported from a gzipped CSV file (`data/blockheaders.csv.gz`)
     - The config options for database preloading haven't changed, except for the different file name (was `blockheaders.xz`, is `blockheaders.csv.gz`)
     - The whole database setup process (migrations and preloading) is much faster than the old method. In my tests less than 5 seconds, vs. a minute in the old method. The performance issues mentioned in the original description below have been resolved.
     - Headers export from the database to the CSV file can be performed via the `--exportHeaders` CLI command. It updates the file `data/blockheaders.csv.gz`.

### What has not been done in this PR and should be done later on in a separate PR

- Removing from `blockheaders.csv` columns which can be calculated during import
- Adding the calculations for the removed columns
- Adding some kind of abstraction for those steps in the import where the operations are specific to SQLite (`modifySQLitePragmas`, etc.)
- Preparing a plan for regular uploading of the header file
- Running export in separate threads to  improve performance (optional)

---

# ORIGINAL DESCRIPTION

## Background

The essence of this task is to change the way block headers data from the chain is pre-loaded during database initialization.

Currently, the repo contains a compressed file with the complete SQLite database snapshot, which is simply unpacked. This approach will not allow us to add support for PostgreSQL (or other databases), so we need to change it.

## Refactorings

While working on this task, I had to refactor a few things, so I'll start by explaining these refactorings.

1. I moved the CLI commands to a separate package (`cli/cli.go`). Previously they were handled directly in `config/viper.go`, which is not a good place for such functionality.

2. I refactored the database connection and database migrations.

- I moved all the code related to database management to the `database` folder. I created there a `database` package, in which I placed the code for connecting to the database and running migrations. I also moved the existing `sql` package from `data/sql` to database, but without any changes.

- I removed the database handling code from `main.go`. Methods from the `database` package are now called in `main`. Migrations are also called in `main`, as a clearly defined step (previously they were hidden somewhere in the SQLite code)

- I converted the implementation for SQLite into an adapter, we should now be able to easily add implementations for other databases.

Now let's get to the main topic.

## Headers export and import

I have added 2 new CLI commands:

- `--exportHeaders` - export headers from database to CSV file
- `--importHeaders` - import headers from CSV file to database

The export command will export the headers data from the database to `data/headers.csv.gz` and the import command will import them from this file. The import command requires the `headers` table to be empty, I added this as a safety check to prevent accidental import.

The solution sort of works, but there are quite a few problems to consider.

## Problems

1. With the current size of the BSV blockchain the headers file is quite large, when compressed it is over 70 MB.

2. Header import row by row from a file is slow. With the current number of headers it takes more than 2 hours on my machine.

4. Running import and export as CLI commands may not be what we want. CLI commands are convenient when working locally, but they may not be suitable for production servers. 

5. We would like to publish the headers file somewhere, but probably not directly in the repo (like the current SQLite database). So where could we put this file? @dorzepowski had an idea to publish it in the release assets on GitHub. This should work fine, I think. But maybe somebody can see problems with this or has other ideas.

6. We would like to update this file periodically, how could we go about doing this? @dorzepowski suggested creating an endpoint in the Pulse server from which the header file could be downloaded. We would then have a job that would download the file from that endpoint and upload it to where we want to store it (release assets?). This is a good idea, but it also has some problems. That's a lot of data to transfer, so if this endpoint is abused, it could bring down the server. Please let me know your thoughts on this.

7. What about the existing database related config options? Some of them don't make sense with this new approach, and some may have to be changed:

* `dbFilePath` - Currently this is the path to the SQLite file. With a database agnostic approach this configuration option doesn't make much sense, Postgres and other potential databases will not have a file path. But then again, we want to support SQLite as well, so maybe we should have this configuration option, just rename it to `sqliteFilePath`?

* `preparedDb` - With this option Pulse sets up the SQLite database from the compressed snapshot in the repo. As we get rid of this snapshot and offer headers import from a CSV file instead, this option should now trigger the data import. But as mentioned earlier, this is a very slow process which can take some hours to complete. With this in mind, should we keep this option, or get rid of it in favour of the `--importHeaders` CLI command?

* `preparedDbFilePath` - This is the path where the compressed SQLite database file is located. With this file gone it doesn't make any sense to have such option. But maybe we need instead an option for the location of the headers CSV file? However, if this file is not in the repository, but e.g. in the release assets, then I don't know what this option could be, a URL maybe?

* `resetState` - According the description, this option should clear the headers from the database so that synchronization with the chain will start from the beginning. The way it works now is that the existing SQLite database file is deleted, and since Pulse (in another place, not related to this option) configures the SQLite database from the compressed file, if it doesn't exist, the database is re-initialized. I see a few problems with this option:
**A)** it does not do what it says in the description (it replaces the whole database with the snapshot)
**B)** in a database agnostic approach it needs a completely different implementation (probably an SQL query to truncate the table in the database)
**C)** I don't think this action should be handled via a config option. I can't imagine somebody would want to reset the database every time the server starts. A CLI command `--clearHeadersDatabse` would make more sense IMO.
So I suggest we get rid of this option completely or replace it with a CLI command.


kuba-4chain:
Comment to point 2. and 3. (and potentially 6) from Problems:

I think we could remove the functionality of importing the headers completely, and leave it to the potential user of Pulse.

Every database has it's own way of importing CSV files and doing it with the recommended way will be much faster than importing it line by line directly in Pulse. This should happen on a database level, not on Pulse level. Especially considering that we want Pulse to be database agnostic, which this PR is greatly moving us towards.

So all that we should do, in my opinion, is give the user of Pulse ability to download pre-loaded headers in a format that is easily uploadable to any db and then only care about the Connection URL to the database (or a file path in case of SQLite). The user can then manually import the pre-loaded headers to the db of his choice (from the ones which we will support) and then enjoy quicker sync from Pulse :). For our own comfort, we can write a simple script that would import it to Postgres and/or SQLite.

As for how to periodically update the file, I have no good ideas how to do it. I was thinking about periodically exporting that file from inside Pulse (configurable) and then exposing it as a static file from the server so the GH Actions can download it any time, but it's also leaving the possibility of this static file access being abused.

dbutil/imoport.go
```diff
@@ -0,0 +1,175 @@
...
+	"math/big"
+	"os"
+	"path"
+	"strconv"
+	"time"
+
+	"github.com/bitcoin-sv/pulse/app/logger"
+	"github.com/bitcoin-sv/pulse/config"
+	"github.com/bitcoin-sv/pulse/database"
+	"github.com/bitcoin-sv/pulse/database/sql"
```
kuba-4chain, line 17: Should there be an underscore?

config/viper.go
```diff
@@ -1,33 +1,38 @@
...
+func Init(lf logging.LoggerFactory) (*Config, *CLI) {
 	viper.AutomaticEnv()
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 	setDefaults()
 
+	fs := &PulseFlagSet{}
+	cli := &CLI{}
+
+	initCliFlags(fs, cli)
+	initFlags(fs)
```
kuba-4chain, line 24: It's a minor, but maybe we could put `initCliFlags()` inside `initFlags()`

dbutil/imoport.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	}
+
+	previousBlockHash := chainhash.Hash{}
+	rowIndex := 0
+
+	for {
+		record, err := reader.Read()
+		if err != nil {
+			if err != io.EOF {
```
dorzepowski, line 110: I think that `!error.Is(...)` should be used instead of `!=`

cmd/main.go
```diff
@@ -47,23 +44,34 @@ func main() {
...
+	if cfg.Db.MigrateDb {
+		log.Info("migrating database")
+		if database.DoMigrations(db, cfg.Db); err != nil {
+			log.Errorf("database migration failed because of error %v", err)
+			os.Exit(1)
+		}
+		log.Info("migrating database completed")
+	} else {
+		log.Info("migrate database set to false, skipping migration")
+	}
```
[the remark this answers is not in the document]
arkadiuszos4chain: I feel the same, migration shouldn't be a part of importing data

config/viper.go
arkadiuszos4chain: And I like idea of two separate functions for this. This makes code more segregated

data/headers.csv.gz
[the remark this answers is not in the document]
mwilkosinski: @dorzepowski I just tried the "File archiver" tool on Ubuntu and I could decompress without any problems, so maybe it's just a Mac issue 😉

dbutil/const.go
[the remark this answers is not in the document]
mwilkosinski: With the other suggested changes implemented, we won't need these `consts` at all.

dbutil/const.go
```diff
@@ -0,0 +1,6 @@
+package dbutil
+
+const (
+	compressedHeadersFilePath = "data/headers.csv.gz"
+	tmpDir                    = ".tmp"
```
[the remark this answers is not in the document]
mwilkosinski: We can get it from the `os`

dbutil/file_utils.go
```diff
@@ -0,0 +1,83 @@
...
+	for {
+		record, err := csvReader.Read()
+		if err != nil {
+			break // Reached the end of the CSV file
+		}
+
+		if err := csvWriter.Write(record); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
mwilkosinski: Good point. We can simplify it using `io.Copy`

cmd/main.go
```diff
@@ -47,23 +44,34 @@ func main() {
 	lf := logger.DefaultLoggerFactory()
 	log := lf.NewLogger("main")
 
-	cfg := config.Init(lf)
+	cfg, cliFlags := config.Init(lf)
+
+	cli.ParseCliFlags(cliFlags, cfg)
 
-	// Unzip prepared db file if configured.
+	// TODO: Should this still be here, or should we support DB import only via CLI commands?
```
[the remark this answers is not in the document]
mwilkosinski: All options 3 are now supported without change.

dbutil/export.go
```diff
@@ -0,0 +1,139 @@
...
+	selectHeadersSql = `
+	SELECT
+		hash,
+		version,
+		merkleroot,
+		nonce,
+		bits,
+		chainwork,
+		strftime('%s', timestamp) as timestamp,
+		cumulatedWork
```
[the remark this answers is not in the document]
mwilkosinski: Calculations should be done in a separate PR

dbutil/export.go
```diff
@@ -0,0 +1,139 @@
...
+
+const (
+	selectHeadersSql = `
+	SELECT
+		hash,
+		version,
+		merkleroot,
+		nonce,
+		bits,
+		chainwork,
```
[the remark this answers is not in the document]
mwilkosinski: Calculations should be done in a separate PR

cli/cli.go
```diff
@@ -0,0 +1,37 @@
...
+		os.Exit(0)
+	}
+
+	if cli.ShowVersion {
+		fmt.Println("pulse", "version", version.String())
+		os.Exit(0)
+	}
+
+	if cli.ExportHeaders {
+		// database.ExportHeaders(cfg, log)
```
Nazarii-4chain, line 30: We can remove this comment

database/export.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// func queryDatabaseTable(db *sqlx.DB, log logging.Logger) *sqlx.Rows {
+// 	rows, err := db.Queryx(selectHeadersSql)
+// 	if err != nil {
+// 		log.Errorf("Failed to query rows: %v", err)
+// 		os.Exit(1)
+// 	}
+// 	return rows
+// }
```
Nazarii-4chain, line 110: We can remove this comment

Nazarii-4chain (dismissed):

Nazarii-4chain (dismissed):
LGTM

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+		return err
+	}
+
+	if err := compressedHeadersFile.Close(); err != nil {
+		return err
+	}
+
+	log.Infof("Decompressed and wrote contents to %s", tmpHeadersFilePath)
+
+	pragmas, err := getSQLitePragmaValues(db)
```
arkadiuszos4chain, line 87: Shouldn't we use db adapter here?

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := modifySQLitePragmas(db); err != nil {
+		return err
+	}
+
+	droppedIndexes, err := removeIndexes(db)
```
arkadiuszos4chain, line 96: Great idea!

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+
+	if err := importHeadersFromFileConcurrent(headersRepo, tmpHeadersFile, log); err != nil {
+		return err
+	}
+
+	if err := tmpHeadersFile.Close(); err != nil {
+		return err
+	}
+
+	if err = restoreIndexes(db, droppedIndexes); err != nil {
```
arkadiuszos4chain, line 109: IMO we should restore indexes even when faced errors - run it as deferred

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+func initHeadersRepo(db *sqlx.DB, cfg *config.Config) *repository.HeaderRepository {
+	lf := logger.DefaultLoggerFactory()
+	headersDb := sql.NewHeadersDb(db, cfg.Db.Type, lf)
+	headersRepo := repository.NewHeadersRepository(headersDb)
+	return headersRepo
+}
+
+func importHeadersFromFileConcurrent(repo repository.Headers, inputFile *os.File, log logging.Logger) error {
```
arkadiuszos4chain, line 133: I am having trouble understanding this code:
1. Errors are placed in the `errCh` channel, but they are never read from the channel.
2. There is no indication of how we wait for workers. I think we should and if we do - how is it done?
3. What does `insertTransactionSize` do? My understanding is that it should represent a bundle/batch size. However, from the code, it seems like we use just `numWorkers` numbers of bundles with non-deterministic sizes and we should use `headersCount / bundleSize` numbers of bundles 

It's possible that my knowledge about go and go routines is not complete to understand this so please correct me.

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil {
+		return err
+	}
+
+	headersRepo := initHeadersRepo(db, cfg)
+
+	dbHeadersCount, _ := headersRepo.GetHeadersCount()
+
+	if dbHeadersCount > 0 {
+		return fmt.Errorf("database already contains %d block headers, the headers table in the database must be empty", dbHeadersCount)
```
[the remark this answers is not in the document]
mwilkosinski: @dorzepowski good idea. I'm changing it to this:

```
if dbHeadersCount > 0 {
  log.Infof("skipping preloading database from file, database already contains %d block headers", dbHeadersCount)
  return nil
}
```

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+		return err
+	}
+
+	if err = restoreSQLitePragmas(db, *pragmas); err != nil {
+		return err
+	}
+
+	if err := os.Remove(tmpHeadersFilePath); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
mwilkosinski: Having implemented the other changes suggested in code review, I have a better idea how to handle this. I'm moving this cleanup to a defered anonymous function:

```
tmpHeadersFile, tmpHeadersFilePath, err := getHeadersFile(cfg.Db.PreparedDbFilePath, log)
if err != nil {
    return err
}
defer func() {
    _ = tmpHeadersFile.Close()

    if fileExistsAndIsReadable(tmpHeadersFilePath) {
        if err := os.Remove(tmpHeadersFilePath); err == nil {
            log.Infof("Deleted temporary file %s", tmpHeadersFilePath)
        } else {
            log.Warnf("Unable to delete temporary file %s", tmpHeadersFilePath)
        }
    }
}()
```

database/import_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	repo := MockHeaders{db: nil}
+	lf := logger.DefaultLoggerFactory()
+	log := lf.NewLogger("import_test")
+	log.Info("TestImportHeadersFromFile")
+	err = importHeadersFromFile(&repo, tmpfile, log)
+	if err != nil {
+		t.Fatalf("Error importing headers: %v", err)
+	}
```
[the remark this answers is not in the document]
mwilkosinski: Agreed offline to remove the `MigrateDb` flag and simplify this `Init` function:

```
func Init(cfg *config.Config, log logging.Logger) (*sqlx.DB, error) {
    db, err := Connect(cfg.Db)
    if err != nil {
        return nil, err
    }

    if err := DoMigrations(db, cfg.Db); err != nil {
        return nil, err
    }

    if cfg.Db.PreparedDb {
        if err := ImportHeaders(db, cfg, log); err != nil {
            return nil, err
        }
    }

    return db, nil
}
```

database/database.go
```diff
@@ -0,0 +1,79 @@
...
+		log.Infof("%s database completed", dbAction)
+	default:
+		log.Info("migrate database set to false, skipping migration")
+	}
+
+	if cfg.Db.PreparedDb {
+		if err := ImportHeaders(db, cfg, log); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
mwilkosinski: Agreed offline to remove the `MigrateDb` flag and simplify this `Init` function:

```
func Init(cfg *config.Config, log logging.Logger) (*sqlx.DB, error) {
    db, err := Connect(cfg.Db)
    if err != nil {
        return nil, err
    }

    if err := DoMigrations(db, cfg.Db); err != nil {
        return nil, err
    }

    if cfg.Db.PreparedDb {
        if err := ImportHeaders(db, cfg, log); err != nil {
            return nil, err
        }
    }

    return db, nil
}
```

database/import.go
mwilkosinski: @arkadiuszos4chain @dorzepowski I have asked @augustync, the author of this code, to explain this 😄

ac4ch: 1. Errors in errCh channel:

The errors placed in the errCh channel are actually consumed in the worker function. In the worker, if there is an error during the database insertion (repo.AddMultipleHeadersToDatabase), the error is sent to errCh. This is a form of error handling, where the main goroutine can later check for errors by receiving from errCh. It might be more explicit to log or handle these errors in the main goroutine as well.

2. Waiting for Workers:

The function uses a combination of channels (doneCh) and goroutine closure to wait for the workers to complete. The defer statement in the worker function sends a signal to doneCh when the worker completes. The main goroutine waits for each worker to finish by receiving from doneCh in a loop (for i := 0; i < numWorkers; i++). This ensures that the main goroutine doesn't proceed beyond the loop until all workers have completed.

3. insertTransactionSize and Batching:

insertTransactionSize represents the number of records to process in each iteration of the outer loop. It acts as a batch size for reading and processing records from the CSV file. The inner loop (for i := 0; i < insertTransactionSize; i++) processes records in batches, and the size of each batch is determined by this variable.
The idea is to efficiently read and process records in chunks, reducing the frequency of database insertions and potentially improving performance. The main goroutine sends blocks to workers in these batches. The use of insertTransactionSize doesn't necessarily imply that each worker gets a bundle; rather, it's about reading records from the file in batches.

dorzepowski: I was mostly interested in p. 1.
I mean, although we worker could put an error on a channel, this channel isn't read at all. And IMHO actually having this error could lead to inconsistent state of DB, because we could think that we insert all the headers, but actually we didn't because there were some errors and for example 1k of headers weren't inserted.

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+			close(blockCh) // Close the channel to signal workers to finish
+			break
+		}
+	}
+
+	log.Infof("Inserted total of %d rows", rowIndex)
+	return nil
+}
+
+func importHeadersFromFile(repo repository.Headers, inputFile *os.File, log logging.Logger) error {
```
dorzepowski, line 226: We don't use it anymore, aren't we?

database/import.go
```diff
@@ -0,0 +1,413 @@
...
+	if err = restoreSQLitePragmas(db, *pragmas); err != nil {
+		return err
+	}
+
+	if err := os.Remove(tmpHeadersFilePath); err != nil {
+		return err
+	}
+
+	log.Infof("Deleted temporary file %s", tmpHeadersFilePath)
+
```
dorzepowski, line 142: And here should go some db consistency check.
We could merge this commit without it, but please add the task/subtask for that. Because when we switch to lighter file (without calculated data) this would be even more important to make the check at the end.

database/import.go
ac4ch: yes that make sense .. that why we can log or handle it better :) feel free to experiment

database/export.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	tmpCsvFile, err := os.Create(tmpHeadersFilePath)
+	if err != nil {
+		return err
+	}
+
+	writer := csv.NewWriter(tmpCsvFile)
+
+	rows, err := queryDatabaseTable(db, log)
```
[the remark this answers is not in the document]
arkadiuszos4chain: Good for me ;-)

arkadiuszos4chain (dismissed):

arkadiuszos4chain (dismissed):

Nazarii-4chain (dismissed):

kuba-4chain (dismissed):

dorzepowski (approved):

sirdeggen (approved):
Looks solid, I can’t run from the airplane today to validate everything works as expect, so will follow up.

I want to point out that we may be able to cut down functionality to simplify the application and make it even faster.

PR #174, opened 2024-01-24
fix(BUX-514): fix dumping config

pawellewandowski98:
# Pull Request Checklist

- [ ] 📖 I created my PR using provided  : [CODE_STANDARDS](https://github.com/bitcoin-sv/pulse/blob/main/.github/CODE_STANDARDS.md)
- [ ] 📖 I have read the short Code of Conduct: [CODE_OF_CONDUCT](https://github.com/bitcoin-sv/pulse/blob/main/.github/CODE_OF_CONDUCT.md)

# Description

- fix dumping config. Default values have to be set before checking flags

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"errors"
+	"github.com/bitcoin-sv/pulse/cli"
 	"net/http"
 	"os"
 	"os/signal"
@@ -15,7 +16,6 @@ import (
 
 	"github.com/bitcoin-sv/pulse/logging"
 
-	"github.com/bitcoin-sv/pulse/cli"
 	"github.com/bitcoin-sv/pulse/config"
 	"github.com/bitcoin-sv/pulse/database"
 	"github.com/bitcoin-sv/pulse/notification"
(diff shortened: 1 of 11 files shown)


config/load.go
```diff
@@ -43,19 +38,25 @@ func Load(cfg *AppConfig) (*AppConfig, *zerolog.Logger, error) {
...
-	if err := mapstructure.Decode(GetDefaultAppConfig(defaultLog), &defaultsMap); err != nil {
+	if err := mapstructure.Decode(GetDefaultAppConfig(), &defaultsMap); err != nil {
 		return err
 	}
 
 	for key, value := range defaultsMap {
 		viper.SetDefault(key, value)
 	}
 
+	Lookup = net.LookupIP
```
wregulski, line 53: I would move it to separated method.

cmd/main.go
```diff
@@ -39,15 +39,20 @@ import (
 func main() {
 	defaultLog := logging.GetDefaultLogger()
 
-	defaultConfig := config.GetDefaultAppConfig(defaultLog)
-	if err := cli.LoadFlags(defaultConfig); err != nil {
+	if err := config.SetDefaults(defaultLog); err != nil {
```
wregulski, line 42: Isn't this method called when we do "Load" in this line?
`cfg, log, err := config.Load(defaultCfg)`

wregulski (changes requested):

config/load.go
pawellewandowski98: Yes, I forgot about it...

cmd/main.go
pawellewandowski98: I removed it from `Load` method because it has to be called before `LoadFlags` and `LoadFlags` cannot be called inside config because of import cycle

wregulski: Ok, understand

kuba-4chain: Sorry to unresolve this, but I just have one more idea. What if we move `flags.go` back from `/cli` to `/config`? 

It's a nice idea to put them in the separate package, but I don't see a good enough reason to have cli flags away from config (after all, flags are config-related as well) and it creates unnecessary complications, such as this.

I would much rather have flags.go in the config package and remove `SetDefaults()` from `main.go` and call it in the config. WDYT?

kuba-4chain (approved):
Generally approve, but I just left one comment-idea

cmd/main.go
kuba-4chain: okay, I missed that, nevermind then

kuba-4chain (approved):

wregulski (approved):

PR #180, opened 2024-01-29
fix(BUX-537): RequireAdmin was running handler even though the abort fcn was executed

chris-4chain:
# Pull Request Checklist


transports/http/auth/require_auth.go
```diff
@@ -14,19 +14,27 @@ import (
...
-			if !t.IsAdmin {
-				c.AbortWithStatus(http.StatusUnauthorized)
-			}
-			handler(c)
 		}
 	}
 	return handler
 }
+
+func _validateToken(c *gin.Context) error {
```
dorzepowski, line 27: why this `_` prefix?

chris-4chain: I thought - based on our code base - that we have such convention, that file-local functions have underscore prefix but if I was wrong, I'm ok to rename it to `validateToken`.

dorzepowski: this convention is only in buxorg, because almost everything there was in one packege. Here we tried to have everything in it's package, so please remove `_` 🙂

dorzepowski (approved):

PR #186, opened 2024-02-01
feat(BUX-289): collecting metrics

chris-4chain:
# Pull Request Checklist

This is the first / basic feature for collecting metrics, which includes:
- counting http requests and measure duration histogram
- collecting additional labels: `method`, `path`, `status`, `classification`
- using gin middleware to monitor all requests

Example Grafana Dashboard based on these metrics:

image

### DIFF
--- a/metrics/gin_middleware.go
+++ b/metrics/gin_middleware.go
@@ -4,15 +4,21 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+var notFoundContextKey = "routeNotFound"
+
 func requestMetricsMiddleware() gin.HandlerFunc {
 	if metrics, enabled := Get(); enabled {
 		return func(c *gin.Context) {
 			tracker := metrics.httpRequests.Track(c.Request.Method, c.Request.URL.Path)
 			tracker.Start()
 			defer func() {
-				// note that the status code will be correct only if higher middleware doesn't change the status code;
-				// order of middlewares matters
-				tracker.End(c.Writer.Status())
+				if _, noRoute := c.Get(notFoundContextKey); noRoute {
+					tracker.EndWithNoRoute()
+				} else {
+					// note that the status code will be correct only if higher middleware doesn't change the status code;
+					// order of middlewares matters
+					tracker.End(c.Writer.Status())
+				}
 			}()
 
 			c.Next()
--- a/metrics/gin_middleware.go
+++ b/metrics/gin_middleware.go
@@ -10,7 +10,7 @@ func requestMetricsMiddleware() gin.HandlerFunc {
 			tracker := metrics.httpRequests.Track(c.Request.Method, c.Request.URL.Path)
 			tracker.Start()
 			defer func() {
-				// note that the status code will be correct only if another middleware doesn't change the status code;
+				// note that the status code will be correct only if higher middleware doesn't change the status code;
 				// order of middlewares matters
 				tracker.End(c.Writer.Status())
 			}()
--- a/metrics/gin_middleware.go
+++ b/metrics/gin_middleware.go
@@ -9,7 +9,11 @@ func requestMetricsMiddleware() gin.HandlerFunc {
 		return func(c *gin.Context) {
 			tracker := metrics.httpRequests.Track(c.Request.Method, c.Request.URL.Path)
 			tracker.Start()
-			defer tracker.End(c.Writer.Status()) //note that the status code will be correct only if another middleware doesn't change the status code; order of middlewares matters
+			defer func() {
+				// note that the status code will be correct only if another middleware doesn't change the status code;
+				// order of middlewares matters
+				tracker.End(c.Writer.Status())
+			}()
 
 			c.Next()
 		}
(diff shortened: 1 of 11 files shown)


Nazarii-4chain (approved):

metrics/request.go
```diff
@@ -0,0 +1,59 @@
...
+	startTime time.Time
+	metrics   *RequestMetrics
+}
+
+func (r *RequestTracker) Start() {
+	r.startTime = time.Now()
+}
+
+func (r *RequestTracker) End(status int) {
+	if status == 404 {
```
dorzepowski, line 45: But what it the endpoint is returning 404 not found because header is not found or something?

metrics/naming.go
```diff
@@ -0,0 +1,16 @@
+package metrics
+
+const serviceName = "pulse"
+const requestsMetricBaseName = "requests"
+
+func metricName(name string) string {
+	return serviceName + "_" + name
```
dorzepowski, line 7: Maybe it's better to use some more generic name for the http metrics and just add label `app` to differ between applications.

dorzepowski (changes requested):

metrics/naming.go
chris-4chain: I hesitated between these two solutions... But I don't see any reason why I shouldn't not do it your way, if you find it beneficial. 
Check my new changes. Below it the example of how one metric-line would look like:
```
bux_requests_total{app="pulse",classification="success",method="GET",path="/some_path",status="200"} 4
```

metrics/request.go
chris-4chain: I found a solution involving custom `gin.NoRoute` handler. So now the metrics-middleware can distinguish between "no-route" 404 from other 404s. 

If you don't like this new solution, I'm afraid that other option is to just delete this safeguard and accept such attack vector...

metrics/gin_middleware.go
```diff
@@ -4,15 +4,21 @@ import (
...
 )
 
+var notFoundContextKey = "routeNotFound"
+
 func requestMetricsMiddleware() gin.HandlerFunc {
 	if metrics, enabled := Get(); enabled {
 		return func(c *gin.Context) {
 			tracker := metrics.httpRequests.Track(c.Request.Method, c.Request.URL.Path)
 			tracker.Start()
 			defer func() {
```
dorzepowski, line 14: one thing just came to me, do we really need this defer?
why can't we just put that code at the end of the function?

metrics/naming.go
```diff
@@ -1,10 +1,10 @@
 package metrics
 
-const serviceName = "pulse"
+const appName = "pulse"
 const requestsMetricBaseName = "requests"
 
 func metricName(name string) string {
-	return serviceName + "_" + name
+	return "bux_" + name
```
dorzepowski, line 7: maybe this prefixing is not needed?
maybe it should only be `http_requests*`
and if other metrics will came, then again we will decide on creation of it, how to name it?

dorzepowski (changes requested):

metrics/gin_middleware.go
chris-4chain: One benefit I can see - of using defer - is: when a next handler throw panic, the defer func with `tracker.End` will be executed.

metrics/naming.go
chris-4chain: I wanted our (bux-toolkit) metrics to be quickly searchable by this prefix (in Grafana for example). 
For example here they added `api_` prefix here: https://prometheus.io/docs/practices/naming/#labels

metrics/gin_middleware.go
dorzepowski: ok, didn't thought about that.

metrics/naming.go
dorzepowski: From the same link:
>...should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. For metrics specific to an application, the prefix is usually the application name itself. **Sometimes, however, metrics are more generic, like standardized metrics exported** by client libraries. Examples:
> - prometheus_notifications_total (specific to the Prometheus server)
> - process_cpu_seconds_total (exported by many client libraries)
> - **http_request_duration_seconds (for all HTTP requests)**

and this is why I'm talking about that, because often I saw `http_request_duration_seconds` as metric name in different systems I was participating in development, therefore it is for me a standard name for that metric.

chris-4chain: So, maybe a hybrid approach: 

- For common metrics in your bux-toolkit, adopt standard naming conventions (e.g., `http_request_duration_seconds`) to ensure ease of understanding 
- Use a custom `bux_` prefix for unique, system-specific metrics to clearly identify them and avoid name collisions. For example `bux_block_height_gauge`.

dorzepowski: Yes, this is something that I even thought at the beginning, I just focused on those http_requests now, because this was the first step.

dorzepowski (approved):

PR #192, opened 2024-02-03
fix(BUX-550): replace wrong character ? in delete methods to make it work with postgres

wregulski:
# Pull Request Checklist

### DIFF
--- a/database/sql/webhooks.go
+++ b/database/sql/webhooks.go
@@ -29,7 +29,7 @@ const (
 
 	sqlDeleteWebhookByUrl = `
 	DELETE FROM webhooks
-	WHERE url = ?
+	WHERE url = :url
 	`
 
 	sqlUpdateWebhook = `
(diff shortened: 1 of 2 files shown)


database/sql/webhooks.go
```diff
@@ -29,7 +29,7 @@ const (
 
 	sqlDeleteWebhookByUrl = `
 	DELETE FROM webhooks
-	WHERE url = ?
```
chris-4chain, line 32: I can see two methods of placeholdering the params - named and with-`?`. Besides inconsistency, I'm wondering why you've changed the `?` to named parameter (`url`) in `sqlDeleteWebhookByUrl` but in the same time you've left, for instance, the `sqlGetWebhookByUrl` unchanged - which looks quite similar to me.

wregulski: @chris-4chain - answer is simple, the rest endpoints are working properly. And while resolving bugs not doing improvements I try to follow minimal-change approach to make it easier to test as is't required for v1.

Inconsistency was there before:
- for insert methods there were params with :something but for other endpoints it was ?

I can change it everywhere if needed, no problem.

chris-4chain: If it works - don't change it 😄

chris-4chain (approved):

PR #203, opened 2024-02-08
fix(BUX-551): redundant methods and webhooks after duplicated block instert

wregulski:
# Pull Request Checklist

### DIFF
--- a/database/sql/headers.go
+++ b/database/sql/headers.go
@@ -17,8 +17,6 @@ import (
 )
 
 const (
-	sqlite3DriverName           = "sqlite3"
-	postgresDriverName          = "postgres"
 	postgresUniqueViolationCode = "23505"
 )
 
@@ -195,7 +193,7 @@ func (h *HeadersDb) Create(ctx context.Context, req dto.DbBlockHeader) error {
 		_ = tx.Rollback()
 	}()
 	if _, err := tx.NamedExecContext(ctx, sqlInsertHeader, req); err != nil {
-		if h.isUniqueViolation(err) {
+		if isUniqueViolation(err) {
 			return customErrs.NewUniqueViolationError()
 		}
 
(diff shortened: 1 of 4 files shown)


errors/unique_violation.go
```diff
@@ -0,0 +1,22 @@
+package errors
+
+type UniqueViolationError struct {
```
wregulski, line 3: If you, my dear reviewer, have another idea on how to pass such an error between layers in order not to use "magic strings" then I will be very grateful for your suggestion :)

database/sql/headers.go
```diff
@@ -187,11 +195,30 @@ func (h *HeadersDb) Create(ctx context.Context, req dto.DbBlockHeader) error {
...
+		}
+
 		return errors.Wrap(err, "failed to insert header")
 	}
+
 	return errors.Wrap(tx.Commit(), "failed to commit tx")
 }
 
+func (h *HeadersDb) isUniqueViolation(err error) bool {
+	switch h.db.DriverName() {
```
chris-4chain, line 209: Here I'm thinking about type switch between `sqlite3.Error` and `pq.Error`. In this approach, the DriverName would not be necessary.

service/chain_service.go
```diff
@@ -188,6 +189,11 @@ func (cs *chainService) previousHeader(bs *domains.BlockHeaderSource) (*domains.
 func (cs *chainService) insert(h *domains.BlockHeader) (*domains.BlockHeader, error) {
 	err := cs.Repositories.Headers.AddHeaderToDatabase(*h)
 	if err != nil {
+		if errors.Is(err, customErrs.NewUniqueViolationError()) {
```
arkadiuszos4chain, line 192: I don't understand how this code would prevent multiple webhook calls after `insert()`

wregulski: fixed

database/sql/headers.go
wregulski: fixed - replaced switch DriverName with switch err(type)

service/chain_service.go
arkadiuszos4chain: I believe we don't need this check here. Please move warning to Add() method

database/sql/headers.go
```diff
@@ -187,11 +193,27 @@ func (h *HeadersDb) Create(ctx context.Context, req dto.DbBlockHeader) error {
 		_ = tx.Rollback()
 	}()
 	if _, err := tx.NamedExecContext(ctx, sqlInsertHeader, req); err != nil {
+		if isUniqueViolation(err) {
```
arkadiuszos4chain, line 196: Are you sure we will get this error here, not on `tx.Commit()` ?

arkadiuszos4chain (changes requested):

database/sql/headers.go
wregulski: Yes, 100% sure

service/chain_service.go
wregulski: Moved to Add method

arkadiuszos4chain (approved):

chris-4chain (approved):

PR #207, opened 2024-02-19
feat(BUX-605): gin logs & debugPrint handled by zerolog

chris-4chain:
# Pull Request Checklist

This PR addresses several aspects of logging within a Gin application:

- **Internal debugPrint function**: Gin's internal `debugPrint` function `implemented in gin as: (fmt.Fprintf(DefaultWriter, "[GIN-debug] " + format, values...))` has been configured to use our zerolog logger by setting the `DefaultWriter` early in the initialization process.
- **Error Logging**:  The `DefaultErrorWriter` has also been set to use our zerolog logger. This ensures that errors, such as those caught by `gin.Recovery()`, are logged consistently.
- **Request Logging Middleware**: I've created custom Gin middleware to log incoming requests in a structured format using zerolog. This ensures that request data (path, status code, etc.) is captured in well-defined fields for easier analysis.

### How it works:
1. When debug mode is enabled, Gin's startup messages are logged through our zerolog setup (customized `DefaultWriter`).
2. Our custom middleware intercepts each request and generates a structured zerolog record.
3. Gin's standard Recovery middleware now logs errors and panics to zerolog (customized `DefaultErrorWriter`).

### Release Mode: 
I've added a new configuration option release_mode (true/false) to control Gin's debug mode.  Setting this to true is essential for production environments.

Example log record:
image

Info about `DefaultWriter`:
https://github.com/gin-gonic/gin/issues/1877#issuecomment-552637900

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -76,8 +76,6 @@ logging:
   instance_name: block-header-service
   # Flag for enabling origin parameter in logs
   origin: false
-  # Release mode for Gin web framework should be enabled in "production"
-  gin_release_mode: false
 
 # Prometheus metrics configuration
 metrics:
--- a/config/config.go
+++ b/config/config.go
@@ -107,8 +107,6 @@ type HTTPConfig struct {
 	AuthToken string `mapstructure:"auth_token"`
 	// ProfilingEndpointsEnabled is a flag for enabling additional endpoits for profiling with use of pprof.
 	ProfilingEndpointsEnabled bool `mapstructure:"debug_profiling"`
-	// GinReleaseMode is a flag for enabling gin.SetMode(gin.GinReleaseMode)
-	GinReleaseMode bool `mapstructure:"gin_release_mode"`
 }
 
 // P2PConfig represents a p2p config.
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -37,7 +37,6 @@ func getHttpConfigDefaults() *HTTPConfig {
 		UseAuth:                   true,
 		AuthToken:                 "mQZQ6WmxURxWz5ch",
 		ProfilingEndpointsEnabled: true,
-		GinReleaseMode:            false,
 	}
 }
 
--- a/logging/gin_loggers.go
+++ b/logging/gin_loggers.go
@@ -11,7 +11,7 @@ import (
 // and every other events when it uses fmt.Fprint(DefaultWriter/DefaultErrorWriter, ...)
 // https://github.com/gin-gonic/gin/issues/1877#issuecomment-552637900
 func SetGinWriters(log *zerolog.Logger) {
-	gin.DefaultWriter = newGinLogsWriter(log, ginDefaultWriterLevel(log))
+	gin.DefaultWriter = newGinLogsWriter(log, zerolog.DebugLevel)
 	gin.DefaultErrorWriter = newGinLogsWriter(log, zerolog.ErrorLevel)
 }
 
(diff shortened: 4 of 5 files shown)


Nazarii-4chain (approved):

transports/http/server/gin_loggers.go
```diff
@@ -0,0 +1,87 @@
...
+func setGinGlobals(cfg *config.HTTPConfig, log *zerolog.Logger) {
+	if cfg.ReleaseMode {
+		gin.SetMode(gin.ReleaseMode)
+	}
+
+	// Make GIN to use our logger for debugPrint, recovery messages and every other events when it uses fmt.Fprint(DefaultWriter/DefaultErrorWriter, ...)
+	// https://github.com/gin-gonic/gin/issues/1877#issuecomment-552637900
+	gin.DefaultWriter = newGinLogsWriter(log, ginDefaultWriterLevel(log))
+	gin.DefaultErrorWriter = newGinLogsWriter(log, zerolog.ErrorLevel)
+}
```
chris-4chain, line 20: I've realized that the `setGinGlobals` function might not be the best fit within the `gin_loggers.go` file, even though its primary purpose is setting global DefaultWriters. 

Let me know, and I'll explore ways to restructure this.

kuba-4chain: IMHO, it would be nice to have this entire file in `/logging` directory, if it's possible, as it's all about logging

config.example.yaml
```diff
@@ -76,6 +76,8 @@ logging:
   instance_name: block-header-service
   # Flag for enabling origin parameter in logs
   origin: false
+  # Release mode should be enabled in "production"
+  release_mode: false
```
kuba-4chain, line 80: Can we call it `gin_release_mode`? Because it's easy to get confused 😄

transports/http/server/gin_loggers.go
christomecki: I did as you suggested 👍

chris-4chain: I did as you suggested 👍

config.example.yaml
chris-4chain: Yes we can 👍 Please re-review this to ensure it's no longer confusing.

kuba-4chain (approved):

config.example.yaml
```diff
@@ -76,6 +76,8 @@ logging:
   instance_name: block-header-service
   # Flag for enabling origin parameter in logs
   origin: false
+  # Release mode for Gin web framework should be enabled in "production"
+  gin_release_mode: false
```
arkadiuszos4chain, line 80: I wonder if we really need the ability to set `gin_mode` separately from the login level. Maybe it's enough to assume that `log.level == debug => gin_mode == debug`?

kuba-4chain: If I understand correctly, it's not about the log level. It's about production/development. If someone wants to have `log.level: debug` in production, he shouldn't automatically have `gin_mode` in `debug`

arkadiuszos4chain: but if you set gin_mode to Release it will not generate debug informations, on the other hand if you set gin to debug mode but your logger will be set to eg. Info level you won't see them either

chris-4chain: As I looking into the `gin` repo. They use it only `to print/or not to print` internal debug messages:
```golang
func IsDebugging() bool {
	return ginMode == debugCode
}
```
https://github.com/gin-gonic/gin/blob/ecdbbbe9483dd12222f2085f717a2c7cb5ac55fe/debug.go#L19C6-L19C17

So maybe @arkadiuszos4chain has a good point to set gin_release_mode when `logLevel >= INFO` 🤔

arkadiuszos4chain: Let me ask this question differently: do we need the ability to set a dependent module (such as gin) in a different mode than the main application?

chris-4chain: Maybe, when we do not wan't to collect huge amount of logs from entire app but we'd like to "debug" a problem with gin which occurs only at production (with avarage/high load).

But as I'm rethinking this issue, I would agree with @arkadiuszos4chain suggestion

chris-4chain: After a conversation with Arek, I've removed the new config parameter and made the Gin release mode dependent on the log level. Please re-review.

arkadiuszos4chain (approved):

kuba-4chain (approved):

PR #213, opened 2024-02-27
feat(188): improving the way initial headers are preloaded, reducing headers file size

jakubmkowalski:
# Pull Request Checklist

Now the csv file is reduced to the essential fields (version, merkleroot, nonce, bits, timestamp), the other fields stored in the database are calculated during the import process


database/postgresql_adapter.go
```diff
@@ -140,6 +143,8 @@ func (a *postgreSqlAdapter) copyHeaders(reader *csv.Reader, batchSize int, previ
 		return
 	}
 
+	bh := service.DefaultBlockHasher()
```
kuba-4chain, line 146: Why define it here? I think it could be defined `calculateFields` method, where it's being used

database/postgresql_adapter.go
```diff
@@ -155,8 +160,13 @@ func (a *postgreSqlAdapter) copyHeaders(reader *csv.Reader, batchSize int, previ
 		if len(record) == 0 {
 			break
 		}
+		var b dto.DbBlockHeader
+		b, err = PrepareRecord(record, lastBlockHash, bh, cumulatedChainWork, lastRowIndex)
+		if err != nil {
+			fmt.Printf("Error while preparing record: %v", err.Error())
+			os.Exit(1)
```
kuba-4chain, line 167: I think we shouldn't Exit here, because BHS can always get headers from peers from even the very beginning if the import fails. But maybe it's a topic for discussion on the channel

database/sqlite_adapter.go
```diff
@@ -218,10 +222,15 @@ func (a *sqLiteAdapter) insertHeaders(reader *csv.Reader, repo *sql.HeadersDb, b
 		if len(record) == 0 {
 			break
 		}
-
-		block := parseRecord(record, int32(lastRowIndex), lastBlockHash)
+		var block dto.DbBlockHeader
+		block, err = PrepareRecord(record, lastBlockHash, bh, cumulatedChainwork, lastRowIndex)
+		if err != nil {
+			fmt.Printf("Error while preparing record: %v", err.Error())
+			os.Exit(1)
```
kuba-4chain, line 229: Same here with Exit

database/sqlite_adapter.go
```diff
@@ -200,10 +202,12 @@ func (a *sqLiteAdapter) dropTableIndexes(table string) (func() error, error) {
 	return dropIndexes(a.db, &q)
 }
 
-func (a *sqLiteAdapter) insertHeaders(reader *csv.Reader, repo *sql.HeadersDb, batchSize int, previousBlockHash string, rowIndex int) (lastRowIndex int, lastBlockHash string, err error) {
+func (a *sqLiteAdapter) insertHeaders(reader *csv.Reader, repo *sql.HeadersDb, batchSize int, previousBlockHash string, cumulatedLastBlockChainWork string, rowIndex int) (lastRowIndex int, lastBlockHash string, cumulatedChainwork string, err error) {
 	lastRowIndex = rowIndex
 	lastBlockHash = previousBlockHash
 	batch := make([]dto.DbBlockHeader, 0, batchSize)
+	cumulatedChainwork = cumulatedLastBlockChainWork
+	bh := service.DefaultBlockHasher()
```
kuba-4chain, line 210: Same here with `bh`, it could be defined in the method that uses it IMO :)

database/import.go
```diff
@@ -102,44 +104,79 @@ func dropHeadersFile(tmpHeadersFile *os.File, tmpHeadersFilePath string, log *ze
...
-	chainWork := parseBigInt(record[5])
-	timestamp := parseInt64(record[6])
-	cumulatedWork := parseBigInt(record[7])
+func PrepareRecord(record []string, previousBlockHash string, bh service.BlockHasher, cumulatedChainWork string, rowIndex int) (dto.DbBlockHeader, error) {
+	parsedRow, err := parseRecordToBlockHeadersSource(record, previousBlockHash)
+	if err != nil {
+		return dto.DbBlockHeader{}, err
+	}
+	cumulatedChainWorkBigInt := parseBigInt(cumulatedChainWork)
+	rowIndexInt32 := int32(rowIndex)
```
kuba-4chain, line 113: I would put these 2 lines into `calculateFields` method, will be a bit cleaner :)

database/import.go
```diff
@@ -102,44 +104,79 @@ func dropHeadersFile(tmpHeadersFile *os.File, tmpHeadersFilePath string, log *ze
...
+	preparedRecord := calculateFields(bh, parsedRow, cumulatedChainWorkBigInt, rowIndexInt32)
+	return preparedRecord, nil
+}
+
+func parseRecordToBlockHeadersSource(record []string, previousBlockHash string) (domains.BlockHeaderSource, error) {
+	if len(record) != 5 {
```
kuba-4chain, line 119: This '5' should be calculated from the number of columns in the db, or at least a const somewhere on top of the file, to avoid magic numbers

database/import.go
```diff
@@ -102,44 +104,79 @@ func dropHeadersFile(tmpHeadersFile *os.File, tmpHeadersFilePath string, log *ze
...
-func parseRecord(record []string, rowIndex int32, previousBlockHash string) dto.DbBlockHeader {
-	hash := parseChainHash(record[0])
-	version := parseInt(record[1])
-	merkleroot := parseChainHash(record[2])
-	nonce := parseInt(record[3])
-	bits := parseInt(record[4])
-	chainWork := parseBigInt(record[5])
-	timestamp := parseInt64(record[6])
-	cumulatedWork := parseBigInt(record[7])
+func PrepareRecord(record []string, previousBlockHash string, bh service.BlockHasher, cumulatedChainWork string, rowIndex int) (dto.DbBlockHeader, error) {
```
arkadiuszos4chain, line 107: Probably can by unexported function

database/import.go
```diff
@@ -102,44 +104,79 @@ func dropHeadersFile(tmpHeadersFile *os.File, tmpHeadersFilePath string, log *ze
...
+	if err != nil {
+		return dto.DbBlockHeader{}, err
+	}
+	cumulatedChainWorkBigInt := parseBigInt(cumulatedChainWork)
+	rowIndexInt32 := int32(rowIndex)
+	preparedRecord := calculateFields(bh, parsedRow, cumulatedChainWorkBigInt, rowIndexInt32)
+	return preparedRecord, nil
+}
+
+func parseRecordToBlockHeadersSource(record []string, previousBlockHash string) (domains.BlockHeaderSource, error) {
```
arkadiuszos4chain, line 118: Please return pointer not a structure

database/postgresql_adapter.go
arkadiuszos4chain: It looks like dependency, I would move it even higher

database/postgresql_adapter.go
arkadiuszos4chain: IMO we can just return error here.

arkadiuszos4chain (changes requested):

database/postgresql_adapter.go
jakubmkowalski: changed to return error

database/sqlite_adapter.go
jakubmkowalski: changed to return error

database/import.go
jakubmkowalski: moved to calculateFields

database/import.go
jakubmkowalski: added const to store number of columns in csv file

database/import.go
jakubmkowalski: changed to operate on pointers

arkadiuszos4chain (approved):

database/import_test.go
```diff
@@ -0,0 +1,324 @@
...
+		},
+		{
+			testID: 1,
+			name:   "newer block, should return valid block header",
+			data: testCaseData{
+				blockRecord: [][]string{
+					{"536870912", "e9446d4ebeb301aeb5a2f375ac062bf3581269d783362cf066f08bbe6040a885", "3035389718", "403300437", "1708954423"},
+				},
+				previousBlockHash:  "0000000000000000031817e0b646350cac1b8770d6cba60717e86185cadb15cc",
+				blockHasher:        service.DefaultBlockHasher(),
```
dorzepowski: It could be created just in test body, or even be a "const"-like variable

database/import_test.go
```diff
@@ -0,0 +1,324 @@
...
+				t.Errorf("Error while preparing record: %v", err)
+			}
+			result = append(result, *block)
+			tc.data.previousBlockHash = block.Hash
+			tc.data.cumulatedChainWork = block.CumulatedWork
+			tc.data.rowIndex++
+		}
+		assert.Equal[dto.DbBlockHeader](t, result[tc.data.numberOfBlocks-1], *tc.expectedBlock)
+		assert.NoError(t, err)
+	}
```
dorzepowski: This way, it will display result for each test case separately, so if one would fail, we will know which one.
```suggestion
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			var result []dto.DbBlockHeader
			var err error
			for i := 0; i < tc.data.numberOfBlocks; i++ {
				block, err := PrepareRecord(tc.data.blockRecord[i], tc.data.previousBlockHash, tc.data.blockHasher, tc.data.cumulatedChainWork, tc.data.rowIndex)
				if err != nil {
					t.Errorf("Error while preparing record: %v", err)
				}
				result = append(result, *block)
				tc.data.previousBlockHash = block.Hash
				tc.data.cumulatedChainWork = block.CumulatedWork
				tc.data.rowIndex++
			}
			assert.Equal[dto.DbBlockHeader](t, result[tc.data.numberOfBlocks-1], *tc.expectedBlock)
			assert.NoError(t, err)
		})
	}
```

database/import_test.go
```diff
@@ -0,0 +1,324 @@
...
+				blockRecord: [][]string{
+					{"2147483648", "e9446d4ebeb301aeb5a2f375ac062bf3581269d783362cf066f08bbe6040a885", "3035389718", "403300437", "1708954423"},
+				},
+				previousBlockHash:  "0000000000000000031817e0b646350cac1b8770d6cba60717e86185cadb15cc",
+				blockHasher:        service.DefaultBlockHasher(),
+				cumulatedChainWork: "409554438998846785912755332",
+				rowIndex:           833233,
+				numberOfBlocks:     1,
+			},
+			expectedBlock:        nil,
```
dorzepowski, line 158: as we always expect the block to be nil, why to declare it in a test case. It should be just checked in a test.

database/import_test.go
```diff
@@ -0,0 +1,324 @@
...
+			expectedErrorMessage: "max hash string length is 64 bytes",
+		},
+	}
+
+	for _, tc := range testCases {
```
dorzepowski, line 298: similar as above, you should change this to use `t.Run(tc.name, func(t *testing.T) {...})` so we would get the feedback which test is failing

database/import_test.go
```diff
@@ -0,0 +1,324 @@
...
+					{"2147483648", "e9446d4ebeb301aeb5a2f375ac062bf3581269d783362cf066f08bbe6040a885", "3035389718", "403300437", "1708954423"},
+				},
+				previousBlockHash:  "0000000000000000031817e0b646350cac1b8770d6cba60717e86185cadb15cc",
+				blockHasher:        service.DefaultBlockHasher(),
+				cumulatedChainWork: "409554438998846785912755332",
+				rowIndex:           833233,
+				numberOfBlocks:     1,
+			},
+			expectedBlock:        nil,
+			expectedErrorMessage: "strconv.ParseInt: parsing \"2147483648\": value out of range",
```
dorzepowski, line 159: It would be good to have more precise message - actually suggesting which value can't be parsed.

The same comment applies to other test cases

database/import.go
```diff
@@ -212,7 +255,7 @@ func validatePrevHashColumn(db *sqlx.DB) error {
 	}
 
 	if count != 1 {
```
dorzepowski, line 257: You should actually adjust the validation because now we have more power to do it correctly.
IMHO what should we check is:
1. `count = importCount` (already implemented)
2. `maxHeight = importCount - 1` (already implemented)
3. heightUniqueness (already implemented)
4. all previous hashes have match in a hash column, excluding check for genesis parent hash **(not implemented)**
5. checkpoints hashes exists in db at expected height **(not implemented)**

dorzepowski (approved):

database/import.go
dorzepowski: For hashes you can probably use something like
```
select h.hash
from headers h
left join headers ph on h.previous_block = ph.hash
where h.height != 0 and ph.hash is null
```

dorzepowski (changes requested):

database/import.go
dorzepowski: checking checkpoints could be done this way:
```
with checkpoints as
         (select 11111 as height, '0000000069e244f73d78e8fd29ba2fd2ed618bd6fa2ee92559f542fdb26e7c1d' as hash union
          select 33333 as height, '000000002dd5588a74784eaa7ab0507a18ad16a236e7b1ce69f00d7ddfb5d0a6' as hash union
          select 74000 as height, '0000000000573993a3c9e41ce34471c079dcf5f52a0e824a81e7f953b8661a20' as hash union
          select 105000 as height, '00000000000291ce28027faea320c8d2b054b2e0fe44a773f3eefb151d6bdc97' as hash union
          select 134444 as height, '00000000000005b12ffd4cd315cd34ffd4a594f430ac814c91184a0d42d2b0fe' as hash union
          select 168000 as height, '000000000000099e61ea72015e79632f216fe6cb33d7899acb35b75c8303b763' as hash union
          select 193000 as height, '000000000000059f452a5f7340de6682a977387c17010ff6e6c3bd83ca8b1317' as hash union
          select 210000 as height, '000000000000048b95347e83192f69cf0366076336c639f9b7228e9ba171342e' as hash union
          select 216116 as height, '00000000000001b4f4b433e81ee46494af945cf96014816a4e2370f11b23df4e' as hash union
          select 225430 as height, '00000000000001c108384350f74090433e7fcf79a606b8e797f065b130575932' as hash union
          select 250000 as height, '000000000000003887df1f29024b06fc2200b55f8af8f35453d7be294df2d214' as hash union
          select 267300 as height, '000000000000000a83fbd660e918f218bf37edd92b748ad940483c7c116179ac' as hash union
          select 279000 as height, '0000000000000001ae8c72a0b0c301f67e3afca10e819efa9041e458e9bd7e40' as hash union
          select 300255 as height, '0000000000000000162804527c6e9b9f0563a280525f9d08c12041def0a0f3b2' as hash union
          select 319400 as height, '000000000000000021c6052e9becade189495d1c539aa37c58917305fd15f13b' as hash union
          select 343185 as height, '0000000000000000072b8bf361d01a6ba7d445dd024203fafc78768ed4368554' as hash union
          select 352940 as height, '000000000000000010755df42dba556bb72be6a32f3ce0b6941ce4430152c9ff' as hash union
          select 382320 as height, '00000000000000000a8dc6ed5b133d0eb2fd6af56203e4159789b092defd8ab2' as hash union
          select 400000 as height, '000000000000000004ec466ce4732fe6f1ed1cddc2ed4b328fff5224276e3f6f' as hash union
          select 430000 as height, '000000000000000001868b2bb3a285f3cc6b33ea234eb70facf4dcdf22186b87' as hash union
          select 470000 as height, '0000000000000000006c539c722e280a0769abd510af0073430159d71e6d7589' as hash union
          select 510000 as height, '00000000000000000367922b6457e21d591ef86b360d78a598b14c2f1f6b0e04' as hash union
          select 552979 as height, '0000000000000000015648768ac1b788a83187d706f858919fcc5c096b76fbf2' as hash union
          select 556767 as height, '000000000000000001d956714215d96ffc00e0afda4cd0a96c96f8d802b1662b' as hash)
select ch.hash
from checkpoints ch
join headers h on ch.hash = h.hash and ch.height = h.height
where h.hash is null
```

But actually those parts 
```
select xxx as height, 'yyy' as hash
```
should be build dynamically in code with use of Checkpoints list from: internal/chaincfg/params.go:253

dorzepowski: After some brainstorming and discussion actually point 4 & 5 can be replaced only by checking if the latest (highest) checkpoint exist in db, this should ensure that all previous blocks are correct.

kuba-4chain (approved):

database/import.go
jakubmkowalski: changed to private

database/import.go
```diff
@@ -102,44 +108,80 @@ func dropHeadersFile(tmpHeadersFile *os.File, tmpHeadersFilePath string, log *ze
...
-		Version:       int32(version),
-		MerkleRoot:    merkleroot.String(),
-		Timestamp:     time.Unix(timestamp, 0),
-		Bits:          uint32(bits),
-		Nonce:         uint32(nonce),
-		State:         "LONGEST_CHAIN",
-		Chainwork:     chainWork.String(),
-		CumulatedWork: cumulatedWork.String(),
-		PreviousBlock: previousBlockHash,
+func PrepareRecord(record []string, previousBlockHash string, bh service.BlockHasher, cumulatedChainWork string, rowIndex int) (*dto.DbBlockHeader, error) {
```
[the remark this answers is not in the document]
jakubmkowalski: changed to private

database/sqlite_adapter.go
```diff
@@ -200,10 +202,12 @@ func (a *sqLiteAdapter) dropTableIndexes(table string) (func() error, error) {
...
 
 	for i := 0; i < batchSize; i++ {
 		record, readErr := reader.Read()
```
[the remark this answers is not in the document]
jakubmkowalski: readErr changed to variable err

database/import_test.go
jakubmkowalski: Outdated because initialization of service.DefaultBlockHasher() has been moved to the calculateFields function, where it is used, so it does not need to be passed anymore. Tests updated

database/import_test.go
jakubmkowalski: removed

database/import_test.go
jakubmkowalski: Done as well as above

database/import.go
jakubmkowalski: A newer block (height 826638) and some older ones have been added to the checkpoints in params.go. Now when the base is loaded, the hash of the last checkpoint will be compared with the hash of a block of the same height in the database

dorzepowski (changes requested):

database/import.go
jakubmkowalski: removed

dorzepowski (approved):

wregulski:
testing

PR #218, opened 2024-03-05
feat(BUX-626): temporarily ban peers that we can't connect to multiple times

kuba-4chain:
# Pull Request Checklist


transports/p2p/addrmgr/addrmanager.go
```diff
@@ -211,6 +227,42 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress) {
 	a.Log.Trace().Msgf("Added new address %s for a total of %d addresses", addr, a.nTried+a.nNew)
 }
 
+// BanAddress creates a ban for defaultBanTime for a peer that we cannot connect to.
+func (a *AddrManager) BanAddress(addr string) {
+	a.addrBanned[addr] = time.Now().Add(defaultBanTime)
+
+	// remove from addrTried buckets
```
wregulski, line 234: I know that I repeat myself a lot about this, but instead of those two comments I would create two separated private funcs to handle it - it will be more descriptive for me :D

transports/p2p/connmgr/connmanager.go
```diff
@@ -181,12 +186,30 @@ type ConnManager struct {
...
 
-	cfg            Config
-	wg             sync.WaitGroup
-	failedAttempts uint64
-	requests       chan interface{}
-	log            *zerolog.Logger
-	quit           chan struct{}
+	cfg Config
+	wg  sync.WaitGroup
+	// failedAttempts uint64
```
wregulski, line 191: Do we need this comment?

transports/p2p/addrmgr/addrmanager.go
```diff
@@ -211,6 +227,42 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress) {
...
+				v.refs--
+				if v.refs == 0 {
+					a.nNew--
+					delete(a.addrIndex, k)
+				}
+			}
+		}
+	}
+
+	a.Log.Warn().Msgf("Address %s banned!", addr)
```
wregulski, line 263: Maybe let's put a date in this log, something like:
(pseudocode)
`a.Log.Warn().Msgf("Address %s banned until %v!", addr, banDuration)`

wregulski (changes requested):

transports/p2p/addrmgr/addrmanager.go
wregulski: **MINOR**

transports/p2p/addrmgr/addrmanager.go
wregulski: MINOR

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +232,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
+		if c.Addr != nil && c.Addr.String() != "" && cm.cfg.BanAddress != nil {
+			cm.incAddrConnAttempt(c.Addr.String())
+			if cm.getAddrConnAttempt(c.Addr.String()) >= maxFailedAttempts {
+				cm.cfg.BanAddress(c.Addr.String())
+				return
+			}
 		} else {
-			go cm.NewConnReq()
+			cm.globalFailedAttempts++
+			if cm.globalFailedAttempts >= maxFailedAttempts {
```
chris-4chain, line 243: Is this thread safe? I mean - what happend when multiple goroutines try to modify or check the `globalFailedAttempts`?

transports/p2p/addrmgr/addrmanager.go
```diff
@@ -134,6 +135,10 @@ const (
 	// getAddrPercent is the percentage of total addresses known that we
 	// will share with a call to AddressCache.
 	getAddrPercent = 23
+
+	// defaultBantime is the default amount of time for which to ban
+	// a peer that we cannot connect to.
+	defaultBanTime = time.Hour * 24
```
chris-4chain, line 141: I'm wondering if this `defaultBanTime` value can be changed by user or configuration? If not, maybe a name: `banTime` would be better.

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: You're right, done :)

transports/p2p/connmgr/connmanager.go
kuba-4chain: We don't, removed

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: Done :)

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: Changed to `banTime` :)

transports/p2p/connmgr/connmanager.go
kuba-4chain: You're right, it was implemented like that previously. I changed it to be thread safe

wregulski (approved):

chris-4chain (approved):

transports/p2p/addrmgr/addrmanager.go
```diff
@@ -146,6 +151,17 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress) {
 	}
 
 	addr := NetAddressKey(netAddr)
+
+	// If address in banned addresses and its bantime still active, ignore it
+	// otherwise, remove it from banned addresses and add it again
+	if t, ok := a.addrBanned[addr]; ok {
+		if t.After(time.Now()) {
+			a.Log.Debug().Msgf("New address %s still banned, ignoring", addr)
```
dorzepowski, line 159: Wouldn't that message appear too many times? If so, then maybe we can use the approach with [logging.SampledLogger](https://github.com/bitcoin-sv/block-headers-service/blob/abbc8b47902fc4b695f2be7111cf6161d42a5d2d/logging/logging.go#L73)

transports/p2p/addrmgr/addrmanager.go
```diff
@@ -211,6 +227,48 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress) {
 	a.Log.Trace().Msgf("Added new address %s for a total of %d addresses", addr, a.nTried+a.nNew)
 }
 
+// BanAddress creates a ban for defaultBanTime for a peer that we cannot connect to.
+func (a *AddrManager) BanAddress(addr string) {
```
dorzepowski, line 231: Does it need to be public?

transports/p2p/connmgr/connmanager.go
```diff
@@ -272,7 +323,10 @@ out:
 				conns[connReq.id] = connReq
 				cm.log.Debug().Msgf("Connected to %v", connReq)
 				connReq.retryCount = 0
-				cm.failedAttempts = 0
+				cm.resetGlobalFailedAttempts()
+				if connReq.Addr != nil && connReq.Addr.String() != "" {
+					cm.failedAttempts[connReq.Addr.String()] = 0
+				}
```
dorzepowski, line 329: ### Matter of taste
I would probably extract it into something like:
```suggestion
				cm.resetFailedAttempts(connReq)
```
And inside we can reset global and address failed attempts

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +253,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
 	} else if cm.cfg.GetNewAddress != nil {
-		cm.failedAttempts++
-		if cm.failedAttempts >= maxFailedAttempts {
-			cm.log.Debug().Msgf("Max failed connection attempts reached: [%d] "+
-				"-- retrying connection in: %v", maxFailedAttempts,
-				cm.cfg.RetryDuration)
-			time.AfterFunc(cm.cfg.RetryDuration, func() {
-				cm.NewConnReq()
-			})
+		if c.Addr != nil && c.Addr.String() != "" && cm.cfg.BanAddress != nil {
```
dorzepowski, line 256: When actually c.Addr can be nil or ""?

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +253,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
+			cm.incAddrConnAttempt(c.Addr.String())
+			if cm.getAddrConnAttempt(c.Addr.String()) >= maxFailedAttempts {
+				cm.cfg.BanAddress(c.Addr.String())
+				return
+			}
```
dorzepowski, line 261: It is nested too much, consider extracting a method like: `cm.registerFailedConnectionTo(c)`

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +253,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
+			cm.incGlobalFailedAttempts()
+			if cm.getGlobalFailedAttempts() >= maxFailedAttempts {
+				cm.log.Debug().Msgf("Max failed connection attempts reached: [%d] "+
+					"-- retrying connection in: %v", maxFailedAttempts,
+					cm.cfg.RetryDuration)
+				time.AfterFunc(cm.cfg.RetryDuration, func() {
+					cm.NewConnReq()
+				})
+				return
+			}
```
dorzepowski, line 272: It's nested too much. Consider extracting a method like `cm.registerFailedConnection()`

transports/p2p/connmgr/connmanager.go
```diff
@@ -181,12 +186,51 @@ type ConnManager struct {
...
+}
+
+func (cm *ConnManager) incAddrConnAttempt(addr string) {
+	cm.failedAttemptsMutex.Lock()
+	defer cm.failedAttemptsMutex.Unlock()
+
+	cm.failedAttempts[addr]++
+}
+
+func (cm *ConnManager) getAddrConnAttempt(addr string) uint16 {
```
dorzepowski, line 229: Maybe instead it would be better to have `isAddressConnectionAttemptsExceeded`?

transports/p2p/connmgr/connmanager.go
```diff
@@ -181,12 +186,51 @@ type ConnManager struct {
...
+}
+
+func (cm *ConnManager) resetGlobalFailedAttempts() {
+	cm.globalFailedAttemptsMutex.Lock()
+	defer cm.globalFailedAttemptsMutex.Unlock()
+
+	cm.globalFailedAttempts = 0
+}
+
+func (cm *ConnManager) getGlobalFailedAttempts() uint64 {
```
dorzepowski, line 215: Maybe instead it would be better to have `isGlobalConnectionAttemptsExceeded`?

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +253,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
-		cm.failedAttempts++
-		if cm.failedAttempts >= maxFailedAttempts {
-			cm.log.Debug().Msgf("Max failed connection attempts reached: [%d] "+
-				"-- retrying connection in: %v", maxFailedAttempts,
-				cm.cfg.RetryDuration)
-			time.AfterFunc(cm.cfg.RetryDuration, func() {
-				cm.NewConnReq()
-			})
+		if c.Addr != nil && c.Addr.String() != "" && cm.cfg.BanAddress != nil {
+			cm.incAddrConnAttempt(c.Addr.String())
```
dorzepowski, line 257: Why we don't increment global connection fails here also?

transports/p2p/connmgr/connmanager.go
```diff
@@ -209,17 +253,25 @@ func (cm *ConnManager) handleFailedConn(c *ConnReq) {
...
+			if cm.getAddrConnAttempt(c.Addr.String()) >= maxFailedAttempts {
+				cm.cfg.BanAddress(c.Addr.String())
```
chris-4chain, line 259: I'm worrying that this function `BanAddress` could be used by multiple goroutines and it is not thread safe because it utilizes a standard `map`.

Maybe my concerns are pointless, if so - the function will work completely fine.

transports/p2p/addrmgr/addrmanager.go
```diff
@@ -211,6 +227,48 @@ func (a *AddrManager) updateAddress(netAddr, srcAddr *wire.NetAddress) {
...
+	a.addrBanned[addr] = banTimeUntil
+
+	a.removeAddrFromTried(addr)
+	a.removeAddrFromNew(addr)
+
+	a.Log.Warn().Msgf("Address %s banned until %s!", addr, banTimeUntil.Format(time.RFC1123))
+}
+
+func (a *AddrManager) removeAddrFromTried(addr string) {
+	for _, bucket := range a.addrTried {
```
dorzepowski, line 242: #### only comment for the future (resolve when you read it)

This is soooooo hard to read because of used structure, that I would consider those addrTried and addrNew things that could be somewhere at the top of the list of candidates to refactor.

dorzepowski (changes requested):

transports/p2p/connmgr/connmanager.go
kuba-4chain: Okay, I used the AddrManager mutex, so it should be fine now

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: Not really, it only happens when a peer sends us NewPeer message, which is actually very rarely.

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: We pass this function to connManager in `tranports/p2p/server.go` line 1441, using the server's addrManager, so I think it needs to be exported publicly. Also for testing purposes.

transports/p2p/connmgr/connmanager.go
kuba-4chain: Good point, I changed it so I could actually reuse a mutex and remove one redundant method

transports/p2p/connmgr/connmanager.go
kuba-4chain: In some occasions we get a connection request without an address. For example when the service starts and the peers are not seeded yet, but also sometimes when the app is running for some time. When this happens, it does a nil panic, that's why I implemented this check.

Resolving this is more in a scope of refactor than this task IMO

transports/p2p/connmgr/connmanager.go
kuba-4chain: Done :)

transports/p2p/connmgr/connmanager.go
kuba-4chain: Done

transports/p2p/connmgr/connmanager.go
kuba-4chain: Done

transports/p2p/connmgr/connmanager.go
kuba-4chain: From my understanding of the code, the global connection failure counter was there because of the connection requests without addresses (that I mentioned before). Without it we have an almost endless loop of connecting until a valid address pops up. But when it comes to connection requests with addresses, there's no need to slow it down, because after 25 unsuccessful attempts, they're gonna be banned anyways (with this solution) and we don't need to block the entire connManager. But if you feel like we should, I can change it :)

transports/p2p/addrmgr/addrmanager.go
kuba-4chain: 100% agree. The difference in structures between addrNew and addrTried and also the address buckets are driving me craazy, they introduce soo much complexity without any significant performance/usefulness benefits.

transports/p2p/connmgr/connmanager.go
dorzepowski: ok, lets leave it this way, at least until we find global solution for those empty addresses

dorzepowski (approved):

PR #219, opened 2024-03-07
chore(BUX-581): update swagger

pawellewandowski98:
# Pull Request Checklist

# Description

- Removed swagger comments from model which is not returned from any http endpoint
- Updated swagger

### DIFF
--- a/domains/headers.go
+++ b/domains/headers.go
@@ -36,8 +36,8 @@ type BlockHeader struct {
 	Bits          uint32         `json:"-"`
 	Nonce         uint32         `json:"nonce"`
 	State         HeaderState    `json:"-"`
-	Chainwork     *big.Int       `json:"-"                 swaggertype:"string"`
-	CumulatedWork *big.Int       `json:"work"              swaggertype:"string"`
+	Chainwork     *big.Int       `json:"-"`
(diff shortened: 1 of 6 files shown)


domains/headers.go
```diff
@@ -36,8 +36,8 @@ type BlockHeader struct {
 	Bits          uint32         `json:"-"`
 	Nonce         uint32         `json:"nonce"`
 	State         HeaderState    `json:"-"`
-	Chainwork     *big.Int       `json:"-"                 swaggertype:"string"`
-	CumulatedWork *big.Int       `json:"work"              swaggertype:"string"`
+	Chainwork     *big.Int       `json:"-"`
+	CumulatedWork *big.Int       `json:"work"`
```
dorzepowski, line 40: Actually it should be string, we should probably change it on endpoints somehow, or maybe there is a method for Json encoder to have custom converter for big.Int

dorzepowski: Yes we should return strings from endpoints, so it would be easier to read and will left for consumer the responsibility for parsing it correctly if he has any needs for that field.

pawellewandowski98: Done

dorzepowski (approved):

Nazarii-4chain (approved):

PR #220, opened 2024-03-08
feat(BUX-631): prevent unneeded header sync

kuba-4chain:
# Pull Request Checklist


chris-4chain (approved):

Nazarii-4chain (approved):

wregulski (approved):

service/chain_service.go
```diff
@@ -97,6 +97,11 @@ func (cs *chainService) Add(bs domains.BlockHeaderSource) (*domains.BlockHeader,
 		return nil, err
 	}
 
+	currentTip, errTip := cs.Headers.GetTip()
```
dorzepowski, line 100: As we are calculating hash in the first line of this method, why not just check if there isn't already header with the same hash stored in database. If there is, then for example return an error like: `header already in database`

dorzepowski (changes requested):

transports/p2p/p2psync/manager.go
```diff
@@ -602,11 +603,16 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
 		return
 	}
 
+	// If we received only blocks that we already have, or are not from the longest chain,
```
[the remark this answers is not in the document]
kuba-4chain: I am checking it, but not directly. So, in this file, inside the for loop, in line 565, if the header is from the longest chain, the `finalHeight` variable is being set to its height. Otherwise it's set to 0. And 0 is also always lower than the current longest chain, so this if-statement also includes this check indirectly.

Maybe it should be refactored, bo for now I tried to do it the minimal way, to not mess up with the existing code too much.

transports/p2p/p2psync/manager.go
```diff
@@ -606,7 +609,6 @@ func (sm *SyncManager) handleHeadersMsg(hmsg *headersMsg) {
 	// headers starting from the latest known header and ending with the
 	// next checkpoint.
 	sm.sendGetHeadersWithPassedParams([]*chainhash.Hash{finalHash}, sm.nextCheckpoint.Hash, peer)
```
chris-4chain, line 611: After last changes the app throws panic (after few seconds of working).

It is related to `finalHash`.

image

image

dorzepowski (approved):

transports/p2p/p2psync/manager.go
dorzepowski: fixed

PR #223, opened 2024-03-13
fix(BUX-649): sending verack after receiving version message

dorzepowski:
# Pull Request Checklist


transports/p2p/server.go
```diff
@@ -509,7 +509,7 @@ func (sp *serverPeer) OnAddr(_ *peer.Peer, msg *wire.MsgAddr) {
 
 		// Add address to known addresses for this peer.
 		sp.addKnownAddresses([]*wire.NetAddress{na})
-		sp.log.Debug().Msgf("Added address %s from host %s", na.IP.String(), sp.NA().IP.String())
+		sp.log.Trace().Msgf("Added address %s from host %s", na.IP.String(), sp.NA().IP.String())
```
dorzepowski, line 512: We're now receiving more then 600 addresses from one seed-peer and we're connecting to 8 of them. So this log is pretty long.

transports/p2p/peer/peer.go
```diff
@@ -1809,6 +1808,32 @@ func (p *Peer) Disconnect() {
 	close(p.quit)
 }
 
+func (p *Peer) readVerOrVerAck() error {
```
chris-4chain, line 1811: I guess, the main purpose of this function is to check for potential error - so consider function name like: "checkVerOrVerAck". 

The same suggestion, concerning `readVerAckMsg` function

arkadiuszos4chain (approved):

wregulski (approved):

transports/p2p/peer/peer.go
dorzepowski: Main purpose is to read Version or VerAck message from p2p "wire" (I don't know how to name it).
Because after sending Version from our side, we can receive either Version of VerAck message, and it seems that for us it is the best to accept both situations because we then can connect more peers.

chris-4chain: Ok, i get it 👍

chris-4chain (approved):

PR #232, opened 2024-04-02
feat(BUX-313): experimental p2p initial implementation

kuba-4chain:
# Pull Request Checklist

# Important Info
What has been done in this PR:
1. Seeding IP's of peers from DNS seed(s)
2. Creating an outbound connection to one peer.
3. Negotiating protocol.
4. Handling ping-pongs to keep the connection active on both sides.
5. Handling of messages: headers, inv, ping, pong
6. Handling of writing messages in a queue
7. Requesting headers from the peer
8. Handling of errors when adding new headers to db

# Keep in mind when checking the code
- I tried to avoid comments, but I left some of them to either explain why something is as it is or to remember about some action in a specific place (TODO comments)
- This is just an initial implementation, a fundament to base upon. If you see where the improvements can be made, please let me know, but don't go too harsh on me 😄 
- There's a lot of logs, for debugging purposes, the final amount of logs will be limited for production.


internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+		select {
+		case <-p.msgChan:
+		default:
+			break cleanup
+		}
+	}
+
+	p.log.Info().Msgf("write msg handler shutdown for peer %s", p)
+	p.wg.Done()
+}
```
chris-4chain, line 302: Personally I don't like the labels 🙈 

Here is how it could look without it:
```golang
func (p *Peer) writeMsgHandler() {
	p.wg.Add(1)
	defer p.wg.Done() // Ensure WaitGroup count is decremented

	for {
		select {
		case msg := <-p.msgChan:
			err := p.writeMessage(msg)
			if err != nil {
				p.log.Error().Msgf("error writing msg %T to peer %s", msg, p)
			}
		case <-p.quit:
			// Drain the channel before exiting
			for msg := range p.msgChan {
				// Optionally handle any remaining messages
			}
			p.log.Info().Msgf("write msg handler shutdown for peer %s", p)
			return // End the goroutine
		}
	}
}
```

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+		return err
+	}
+	return nil
+}
+
+func (p *Peer) startSync() error {
+	p.log.Info().Msgf("requesting headers from peer %s", p)
+
+	var err error
+	if p.nextCheckpoint != nil {
```
chris-4chain, line 143: I'm thinking about concurrency issue here - because the `nextCheckpoint` field is read here (without locking or any other sync method) and "at the same time" it could be modified in the goroutine `readMsgHandler`.

Maybe it's not an issue but I prefer to ask 😄

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+			break cleanup
+		}
+	}
+
+	p.log.Info().Msgf("write msg handler shutdown for peer %s", p)
+	p.wg.Done()
+}
+
+func (p *Peer) writeOurVersionMsg() error {
+	n, err := rand.Int(rand.Reader, big.NewInt(9223372036854775807))
```
chris-4chain, line 305: It looks like a magic number here. Consider making some contant ot add a comment why this actual number.

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	if err != nil {
+		panic(err)
```
chris-4chain, line 307: Why panic instead of returning error?

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	msg := wire.NewMsgVersion(ourNA, theirNA, nonce, lastBlock)
+	err = msg.AddUserAgent(p.cfg.UserAgentName, p.cfg.UserAgentVersion, userAgentComments)
+	if err != nil {
+		p.log.Error().Msgf("could not add user agent to version message, reason: %v", err)
+		return err
+	}
+
+	// NOTE: it's 0 by default, so in theory we don't need
+	// to set that, but it's better to be explicit.
+	msg.Services = p.services
```
chris-4chain, line 329: I can't get the connection between the comment (about "0 by default") and the line below it "msg.Services = p.services"

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+			nonce, err := wire.RandomUint64()
+			if err != nil {
+				p.log.Error().Msgf("error generating nonce for ping msg, reason: %v", err)
+				continue
+			}
+			p.log.Info().Msgf("sending ping to peer %s with nonce: %d", p, nonce)
+			p.queueMessage(wire.NewMsgPing(nonce))
+
+		case <-p.quit:
+			break out
```
chris-4chain, line 202: Consider not using the label "out". 
Instead you can just put here 
```golang
p.log.Info().Msgf("....")
return
```

and after `p.wg.Add` at the begining of the function, add:
```golang
defer p.wg.Done()
```

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+// MsgHandler is a message handler for incoming messages.
+// Must be run as a goroutine.
+func (p *Peer) readMsgHandler() {
+	p.wg.Add(1)
+
+out:
+	for {
+		select {
+		case <-p.quit:
+			break out
```
chris-4chain, line 219: Same here. I would prefer to implement it without the labels. It's only my suggestion 😄  If you really want to utilize the label approach - I'm fine 🙈 
If so, consider renaming the `out` label - because the label should name the code below it. In this case it could look like this:
```golang
mainloop:
for {
		select {
		case <-p.quit:
			break mainloop
                //(...)
}
```

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	}
+}
+
+func (p *Peer) queueMessage(msg wire.Message) {
+	// running in goroutine here, because writing
+	// to channel is blocking until read
+	go func() {
+		p.msgChan <- msg
+	}()
+}
```
chris-4chain, line 271: As far as I understand your code - the `msgChan` is an *unbuffered* channel. So it will not act as a queue.
So the name of the method doesn't fit the implemented behaviour. 
Or I'm wrong 🙈 😄 


https://medium.com/@suryanshshrivastava_75738/unbuffered-channel-is-not-the-same-as-buffered-channel-of-size-1-3065c91344cc

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+
+// writeMsgHandler serves as a queue for writing messages to peers,
+// must be run as a goroutine.
+func (p *Peer) writeMsgHandler() {
+	p.wg.Add(1)
+
+out:
```
chris-4chain, line 278: The label approach :) please consider not xD

internal/transports/p2p/peer/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	for i := len(checkpoints) - 2; i >= 0; i-- {
+		if height >= checkpoints[i].Height {
+			break
+		}
+		nextCheckpoint = &checkpoints[i]
+	}
+	return nextCheckpoint
+}
+
+func searchForFinalBlock(invVects []*wire.InvVect) int {
```
chris-4chain, line 35: Consider add the name `searchForFinalBlockIndex` with `Index` suffix which describes what is returned.

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+
+		if service.HeaderCreationFail.Is(addErr) {
+			p.log.Error().Msgf("couldn't create header from %v because of error %+v", header, addErr)
+			continue
+		}
+
+		if service.ChainUpdateFail.Is(addErr) {
+			p.log.Error().Msgf("when adding header %v couldn't update chains state because of error %+v", header, addErr)
+			continue
+		}
```
chris-4chain, line 484: Consider wrap this code into:
```golang
if addErr != nil {
   //your code with .Is(addErr) checking
}
```

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	)
+	err := p.writeGetHeadersMsg(&zeroHash)
+	if err != nil {
+		p.log.Error().Msgf("error requesting headers from peer %s, reason: %v", p, err)
+	}
+}
+
+func (p *Peer) verifyCheckpointReached(h *domains.BlockHeader, receivedCheckpoint bool) (bool, error) {
+	if p.nextCheckpoint != nil && h.Height == p.nextCheckpoint.Height {
+		if h.Hash == *p.nextCheckpoint.Hash {
```
chris-4chain, line 539: Are you sure you won't get the `null pointer panic` here?

internal/transports/p2p/peer/utils.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/bitcoin-sv/block-headers-service/internal/chaincfg"
+	"github.com/bitcoin-sv/block-headers-service/internal/wire"
+)
+
+func minUint32(a, b uint32) uint32 {
+	if a < b {
+		return a
+	}
+	return b
+}
```
chris-4chain, line 13: As far as I recall - golang introduced similar function.
https://www.linkedin.com/pulse/built-in-min-max-functions-go-121-robert-gogolok/

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	addr string,
+	cfg *config.P2PConfig,
+	chainParams *chaincfg.Params,
+	headersService service.Headers,
+	chainService service.Chains,
+	log *zerolog.Logger,
+) (*Peer, error) {
+	port, err := strconv.Atoi(config.ActiveNetParams.DefaultPort)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
kuba-4chain: Done

internal/transports/p2p/peer/peer.go
kuba-4chain: Thanks for the idea! I actually wanted to avoid labels but I didn't know why. There's one more thing about this solution that won't work, but let's move this conversation to priv :)

internal/transports/p2p/peer/peer.go
kuba-4chain: You're right, removed this section completely. I used the much simpler function from `math.rand` `rand.Uint64` for this task

internal/transports/p2p/peer/peer.go
kuba-4chain: Removed

internal/transports/p2p/peer/peer.go
kuba-4chain: Because p.services is 0, and msg.Services is also 0 by default, so we don't need to set this line `msg.Services = p.services`, but I actually want to do that to be explicit in that case.

But you're right that this comment is misleading, removed :)

internal/transports/p2p/peer/peer.go
kuba-4chain: Absolutely,  done :)

internal/transports/p2p/peer/peer.go
kuba-4chain: Done!

internal/transports/p2p/peer/peer.go
kuba-4chain: Sure

internal/transports/p2p/peer/utils.go
kuba-4chain: You're right, done

internal/transports/p2p/peer/peer.go
kuba-4chain: Done

internal/transports/p2p/peer/utils.go
kuba-4chain: I thought it's not gonna work with uint32, but it does, thanks :)

internal/transports/p2p/peer/peer.go
kuba-4chain: Update, the use of `math/rand` created a security comment from golang linter and suggested to use `crypto/rand`, so I did and explained it better, as you suggested

internal/transports/p2p/peer/peer.go
kuba-4chain: Update: returning a formatted error

internal/transports/p2p/peer/peer.go
chris-4chain: Got it 👍

internal/transports/p2p/peer/peer.go
kuba-4chain: Switched to buffered channel. I decided to make a buffer with size 10 for now, because:

- we only send ping and getheaders msg, so 10 should be more than enough
- it's a buffer per peer, so every peer will have 10-msg size queue

internal/transports/p2p/peer/peer.go
kuba-4chain: Safety check added :)

chris-4chain (approved):

dorzepowski (changes requested):

internal/transports/p2p/seed.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/bitcoin-sv/block-headers-service/internal/chaincfg"
+	"github.com/rs/zerolog"
+)
+
+func SeedFromDNS(seeds []chaincfg.DNSSeed, log *zerolog.Logger) []net.IP {
```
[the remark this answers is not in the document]
kuba-4chain: Done

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	if err != nil {
+		return err
+	}
+
+	p.wg.Wait()
+	p.log.Info().Msgf("successfully disconnected peer %s", p)
+	return nil
+}
+
+func (p *Peer) Start() error {
```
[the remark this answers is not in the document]
kuba-4chain: Done

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+	err := p.negotiateProtocol()
+	if err != nil {
+		p.log.Error().Msgf("error negotiating protocol with peer %s, reason: %v", p, err)
+		return err
+	}
```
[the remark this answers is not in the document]
kuba-4chain: Makes sense, done

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+
+	err = p.startSync()
+	if err != nil {
+		p.log.Error().Msgf("error starting sync from peer %s, reason: %v", p, err)
+		return err
+	}
+	return nil
+}
+
+func (p *Peer) startSync() error {
```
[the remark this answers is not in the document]
kuba-4chain: okay, done

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+func (p *Peer) writeOurVersionMsg() error {
+	n, err := rand.Int(rand.Reader, big.NewInt(9223372036854775807))
+	if err != nil {
+		panic(err)
+	}
+	nonce := n.Uint64()
+	p.nonce = nonce
+
+	ourNA := &wire.NetAddress{
+		Services:  wire.SFspv,
```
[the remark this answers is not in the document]
kuba-4chain: done, added to consts

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,550 @@
...
+	lastBlock := p.headersService.GetTip().Height
+
+	msg := wire.NewMsgVersion(ourNA, theirNA, nonce, lastBlock)
+	err = msg.AddUserAgent(p.cfg.UserAgentName, p.cfg.UserAgentVersion, userAgentComments)
+	if err != nil {
+		p.log.Error().Msgf("could not add user agent to version message, reason: %v", err)
+		return err
+	}
+
+	msg.Services = p.services
```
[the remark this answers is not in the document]
kuba-4chain: done, added to consts

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,546 @@
...
+	if msg.Nonce == p.nonce {
+		err := errors.New("disconnecting peer connected to self")
+		p.log.Error().Msg(err.Error())
+		return err
+	}
+
+	p.protocolVersion = min(p.protocolVersion, uint32(msg.ProtocolVersion))
+	p.services = msg.Services
+	p.lastBlock = msg.LastBlock
+	p.startingHeight = msg.LastBlock
```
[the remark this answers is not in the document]
kuba-4chain: Not at all for now, removed

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,546 @@
...
+		}
+	}
+}
+
+func (p *Peer) writeMessage(msg wire.Message) error {
+	err := wire.WriteMessage(p.conn, msg, p.protocolVersion, p.chainParams.Net)
+	if err != nil {
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
kuba-4chain: done

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,558 @@
...
+
+	err = p.readVerOrVerAck()
+	if err != nil {
+		return err
+	}
+
+	err = p.readVerOrVerAck()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
kuba-4chain: I had an idea of how to solve it, based on your implementation, please check if it's okay now

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,546 @@
...
+
+	p.log.Info().Msgf("requesting new headers from peer %s", p)
+	err = p.writeGetHeadersMsg(lastBlockHash)
+	if err != nil {
+		p.log.Error().Msgf("error while requesting new headers from peer %s, reason: %v", p, err)
+	}
+}
+
+func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
```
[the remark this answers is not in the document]
kuba-4chain: I moved the last part to a separate function, other than that, I don't really see a good way of refactoring it, other than throwing some comments, which are unncessary and we agreed to not use them. 

For me this code is pretty self-explanatory, but maybe it's because I've been sitting with it (and with the diagrams) for too long 😄

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,554 @@
...
+
+	p.log.Info().Msgf(
+		"checkpoints synced, requesting headers from height %d up to end of chain (zero hash) from peer %s",
+		lastHeight, p,
+	)
+
+	err := p.writeGetHeadersMsg(&zeroHash)
+	if err != nil {
+		p.log.Error().Msgf("error requesting headers from peer %s, reason: %v", p, err)
+	}
```
dorzepowski, line 549: Maybe we can just call `p.requestHeaders()` here (and eventually put the logs in that method.

internal/transports/p2p/peer/peer.go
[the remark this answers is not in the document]
dorzepowski: Now it looks better, but I think that I still have some idea how to make it a little bit more clear.
If you don't mind, I could do a separate PR after merging this one.

dorzepowski (changes requested):

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,554 @@
...
+
+	p.log.Info().Msgf(
+		"checkpoints synced, requesting headers from height %d up to end of chain (zero hash) from peer %s",
+		lastHeight, p,
+	)
+
+	err := p.writeGetHeadersMsg(&zeroHash)
+	if err != nil {
+		p.log.Error().Msgf("error requesting headers from peer %s, reason: %v", p, err)
```
dorzepowski, line 548: We should add a subtask to handle properly errors when we cannot ask for another batch of headers,
because if we're not able to send `GetHeaders` message, then we won't receive anything, and the whole synchronisation will stuck, and we'ld be doomed 🙂 

But we can resolve that later in a separate task

internal/transports/p2p/peer/peer.go
kuba-4chain: I did it, and also refactored further a little bit

internal/transports/p2p/peer/peer.go
kuba-4chain: Okay, I added a TODO comment saying to lower the peer sync score.

BTW. everywhere I added the TODO comment saying "ban peer or lower peer sync score", I was having in mind that we will be disconnecting the peer and switching to another one for headers syncing. According to our idea about having a pool of open connections to peer, but only one (best) peer that we're syncing the headers from.

I will add a task for it, but we first need to implement the pool of connections in the server and we can't do both of these tasks in paralel.

dorzepowski (approved):

internal/transports/p2p/peer/peer.go
dorzepowski: It's ok, 
Previously the error was only logged so, actually we didn't react to it correctly.

internal/transports/p2p/peer/peer.go
```diff
@@ -0,0 +1,544 @@
...
+		quitting:        false,
+		quit:            make(chan struct{}),
+	}
+	return peer, nil
+}
+
+func (p *Peer) Connect() error {
+	conn, err := net.Dial(p.addr.Network(), p.addr.String())
+	if err != nil {
+		p.log.Error().Msgf("error connecting to the peer, reason: %v", err)
```
wregulski, line 96: It would be great to have peer info here also:
```suggestion
		p.log.Error().Msgf("error connecting to the peer %s, reason: %v", p, err)
```

wregulski (approved):

wregulski:
Really good job! 🫶

internal/transports/p2p/peer/peer.go
kuba-4chain: Right, I missed it, done

PR #235, opened 2024-04-05
feat(BUX-713): sendheaders msg implemented

kuba-4chain:
# Pull Request Checklist

# What's happening in this PR
- `sendheaders` message requests from peer to send us new headers directly, without the inv messsage first. This request needs to be sent to peer only once, and from that point he will send us headers if he has a protocol version >= 70012. [Link to sendheaders BIP](https://github.com/bitcoin/bips/blob/master/bip-0130.mediawiki)
- this request should only be sent after we think we're synced with tip of chain, because if we do it while still syncing, we will get a lot of orphaned blocks, because they will not have parents yet
- in `StartHeadersSync` we set `sendHeadersMode` to `false`, because when we start a new sync (for example, because we're choosing another peer to sync from), we should always assume that this peer may be out of sync with the tip of chain

### DIFF
--- a/internal/transports/p2p/peer/peer.go
+++ b/internal/transports/p2p/peer/peer.go
@@ -21,22 +21,23 @@ import (
 )
 
 type Peer struct {
-	conn            net.Conn
-	addr            *net.TCPAddr
-	cfg             *config.P2PConfig
-	chainParams     *chaincfg.Params
-	checkpoints     []chaincfg.Checkpoint
-	nextCheckpoint  *chaincfg.Checkpoint
-	headersService  service.Headers
-	chainService    service.Chains
-	log             *zerolog.Logger
-	services        wire.ServiceFlag
-	protocolVersion uint32
-	nonce           uint64
-	lastBlock       int32
-	timeOffset      int64
-	userAgent       string
-	synced          bool
+	conn              net.Conn
+	addr              *net.TCPAddr
+	cfg               *config.P2PConfig
+	chainParams       *chaincfg.Params
+	checkpoints       []chaincfg.Checkpoint
+	nextCheckpoint    *chaincfg.Checkpoint
+	headersService    service.Headers
+	chainService      service.Chains
+	log               *zerolog.Logger
+	services          wire.ServiceFlag
+	protocolVersion   uint32
+	nonce             uint64
+	lastBlock         int32
+	timeOffset        int64
+	userAgent         string
+	syncedCheckpoints bool
+	sendHeadersMode   bool
 
 	wg       sync.WaitGroup
 	msgChan  chan wire.Message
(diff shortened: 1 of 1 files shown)


Nazarii-4chain (approved):

wregulski:
For the clarification - current flow is (for new database):
-> import headers from db / start synchronization from genesis,
-> synchronize using invMessage and getHeaders
-> stop on tip and switch to sendHeaders if possible?
---
-> then we receive headers via sendHeaders, and we handle it normally?

internal/transports/p2p/peer/peer.go
```diff
@@ -539,6 +553,14 @@ func (p *Peer) verifyCheckpointReached(h *domains.BlockHeader, receivedCheckpoin
 	return receivedCheckpoint, nil
 }
 
+func (p *Peer) switchToSendHeadersMode() {
+	if !p.sendHeadersMode && p.protocolVersion >= wire.SendHeadersVersion {
+		p.log.Info().Msgf("switching to send headers mode - requesting peer %s to send us headers directly instead of inv msg", p)
+		p.queueMessage(wire.NewMsgSendHeaders())
```
wregulski, line 559: Do we handle response to this message normally, via:
```go
case *wire.MsgHeaders:
	p.handleHeadersMsg(msg)
```
?

kuba-4chain:
Not quite, but very similar. The previous flow was:

1. import headers from db / start synchronization from genesis,
2. Ask for headers to the tip of chain -> send `getheaders` msg
3. Receive headers -> get `headers` msg
4. If we're synced -> Wait for `inv` msg
5. If `inv` msg contains hash that we don't have -> request more headers -> send `getheaders` msg
6. Get new headers -> save to db

The new flow gets rid of point 4 and 5, so:

1. import headers from db / start synchronization from genesis,
2. Ask for headers to the tip of chain -> send `getheaders` msg
3. if we're synced -> ask peer to send headers instead of inv -> send `getheaders` msg (only once)
4. Every time peer has new headers -> Get new headers -> save to db

So instead of  get inv -> send getheaders -> get headers, we just get headers 😃 . Please let me know if I didn't explain it well enough.

chris-4chain (approved):
As far as I understand, the `sendHeadersMode` is just a flag to prevent from sending `p.queueMessage(wire.NewMsgSendHeaders())` multiple times. 

Looks good to me 💯

wregulski (approved):

internal/transports/p2p/peer/peer.go
```diff
@@ -501,6 +506,15 @@ func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
 
 	if headersReceived == 0 {
 		p.log.Debug().Msgf("received only existing headers from peer: %s", p)
+		// If we're synced with all checkpoints and received only existing headers
+		// then it likely means that we're synced with tip of chain, so now we can
+		// request the peer to send us headers directly instead of inv messages.
+		//
+		// If we do this while still syncing, we will get a lot of orphaned blocks,
+		// because the new headers sent by peer will not have parents yet.
+		if p.syncedCheckpoints {
```
dorzepowski, line 515: Maybe instead we should consider a flag (or method) that will tell that we're synced with peers tip. 
And how we can assume that we're synced?
- with Version message we're receiving start_height which means what is the height of this peer at the moment of sending a Version message
- everytime we receive INV we know what is the highest hash of this peer.

Having those information we could assume that we're synced when:
- hash from INV is nil and our tip is at the start_height
- hash from INV is not nil and we have such a hash in db

Then we could use that info we could switch to sendheaders as soon as we reach the tip from the peer.

dorzepowski: ⚠️ We should probably also take into account the time when we last time received INV message, if it is more then  for example 1 or 2 minute old, then we should wait for the next block. This way we will be more safe to not omit some header because of the concurency

dorzepowski: This is only suggestio, fill free to do it any way that would be right.
ℹ️ I would probably go with solution like that:
1. add 2 fields to peer: latestHash and latestHeight
2. when receiving version set latestHeight from version
3. when receiving INV, set latestHash from it and remove latestHeight (set it to nil)
4. when we save header then we can check if it's hash is equal to latestHash and if so, then set latesHeight to the header height
5. when checking if we're synced, check if tip is eq to latestHeight - we could also make this kinda cacheable and set the flag (synced) at the peer in step 4 instead.

dorzepowski (changes requested):

internal/transports/p2p/peer/peer.go
```diff
@@ -501,6 +506,15 @@ func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
 
 	if headersReceived == 0 {
 		p.log.Debug().Msgf("received only existing headers from peer: %s", p)
+		// If we're synced with all checkpoints and received only existing headers
+		// then it likely means that we're synced with tip of chain, so now we can
```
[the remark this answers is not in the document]
kuba-4chain: Switched to new flow, this is no longer the case, resolving.

internal/transports/p2p/peer/peer.go
```diff
@@ -501,6 +506,15 @@ func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
 
 	if headersReceived == 0 {
```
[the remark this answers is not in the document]
kuba-4chain: True, switched to a new flow, resolving.

internal/transports/p2p/peer/peer.go
kuba-4chain: That's a great idea. I finally managed to come up with a simple and elegant (IMHO) solution, based on your suggested flow, but also adding couple of safeguards, checks and a mutex to be thread-safe :)

Please check if the solution now is satisfactory :)

dorzepowski (approved):

PR #241, opened 2024-04-12
refactor(BUX-313): refactoring of checkpoints checking

dorzepowski:
# Pull Request Checklist


internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, -1
+	}
+
+	if height >= ch.finalCheckpoint.Height {
+		return nil, -1
+	}
+
+	if ch.checkpoint != nil {
+		index = ch.currentIndex + 1
+		return &ch.checkpoints[index], index
```
chris-4chain, line 126: The `index out of range` is not explicitly checked here - however I suppose it is done by above conditions. Just make sure that it is safe.

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	nextCheckpoint = ch.finalCheckpoint
+	index = len(ch.checkpoints) - 1
+	for i := index - 1; i >= 0; i-- {
+		if height >= ch.checkpoints[i].Height {
+			break
```
chris-4chain, line 133: Consider just returning the current iteration values:
```golang
return &ch.checkpoints[i], i
```

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		nextCheckpoint = &ch.checkpoints[i]
+		index = i
+	}
+	return nextCheckpoint, index
```
chris-4chain, line 138: What if the above "for" hasn't found the `height` value?

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if len(checkpoints) != 0 {
+		ch.finalCheckpoint = &checkpoints[len(checkpoints)-1]
+		ch.next(tipHeight)
+	}
+
+	return ch
+}
+
+func (ch *checkpoint) Height() int32 {
```
kuba-4chain, line 37: Since this is within `peer` package, we can either make these methods private, or move the checkpoint struct to another package within peer to have good encapsulation practice

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+	ch.checkpoint = nil
+	ch.currentIndex = -1
+	ch.next(height)
+}
+
+func (ch *checkpoint) next(height int32) {
+	if len(ch.checkpoints) == 0 {
+		return
+	}
```
kuba-4chain, line 101: We're also checking it in `findNextCheckpoint()`, I would remove this check from here to make the function cleaner

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+	"sync"
+
+	"github.com/bitcoin-sv/block-headers-service/domains"
+	"github.com/bitcoin-sv/block-headers-service/internal/chaincfg"
+	"github.com/bitcoin-sv/block-headers-service/internal/chaincfg/chainhash"
+	"github.com/rs/zerolog"
+)
+
+type checkpoint struct {
+	checkpoint      *chaincfg.Checkpoint
```
kuba-4chain, line 14: Maybe we could consider renaming `checkpoint` -> `currentCheckpoint`? Because I'm constantly confusing it with `checkpoints` 😄

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+// LastReached returns true if the last checkpoint has been reached.
+func (ch *checkpoint) LastReached() bool {
+	ch.lock.RLock()
+	defer ch.lock.RUnlock()
+
+	return ch.checkpoint == nil
+}
+
+// Check checks if the header is valid according to the checkpoint and marks switches to next checkpoint if reached.
+func (ch *checkpoint) Check(header *domains.BlockHeader) error {
```
kuba-4chain, line 60: A minor, but I like the `Verify` name much more than just `Check` for this method 😅 

Or maybe even `VerifyAndUpdate` because it's also updating to the next checkpoint? 🤔

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+// UseTip sets the next checkpoint to the valid checkpoint for given tip height.
+func (ch *checkpoint) UseTip(height int32) {
+	ch.lock.Lock()
+	defer ch.lock.Unlock()
+
+	ch.checkpoint = nil
+	ch.currentIndex = -1
+	ch.next(height)
+}
```
kuba-4chain, line 96: I don't really see the point of using this method. We don't need to reset `checkpoint` and `currentIndex`, because they are immediately set in the `next` method. 
And I think it makes more sense to just add locking to `next` method and use it in `StartSyncHeaders` instead of this method.

internal/transports/p2p/peer/peer.go
```diff
@@ -134,7 +130,7 @@ func (p *Peer) StartHeadersSync() error {
 	go p.readMsgHandler()
 
 	currentTipHeight := p.headersService.GetTipHeight()
-	p.setNextHeaderCheckpoint(currentTipHeight)
+	p.checkpoint.UseTip(currentTipHeight)
```
kuba-4chain, line 133: We can just use `next` method directly here. We would be passing `currentTipHeight` to this method anyways, so it's descriptive enough IMO

internal/transports/p2p/peer/checkpoint_test.go
```diff
@@ -0,0 +1,280 @@
...
+			chckPoint := newCheckpoint(params.checkpoints, params.tipHeight, &testLogger)
+
+			// expect:
+			assert.Equal(t, params.expectedLastReached, chckPoint.LastReached())
+		})
+	}
+
+}
+
+func TestCheckpointUseTip(t *testing.T) {
```
kuba-4chain, line 57: If we remove `UseTip` method, we can just test `next` method here with the same test cases.

internal/transports/p2p/peer/checkpoint_test.go
kuba-4chain, line 1: I just want to say I love the test cases ❤️

internal/transports/p2p/peer/checkpoint.go
dorzepowski: I already fall into that trap.
It's not so obvious and easy.
Because what we're actually searching for is the checkpoint which:
- is higher then given height
- is after the checkpoint that is lower or equal to given height.

For example:
a. Lets say we're looking for next checkpoint for header `2222`
b. For the sake of simplicity, lets say that we have the following list of checkpoints:
    1. `1111`
    2. `2222`
    3. `3333`
    4. `4444`
 c. In this example we should return checkpoint 3 from that function, but
      -  as you see 2, 3 and 4 will match that condition
      - also 3 and 4 would match the condition height > checkpoint
  d. We could try to write the condition in different way, but I'm not sure if it is much clearer
```
	for i := 0; i < len(ch.checkpoints); i++ {
		index = i + 1
		nextCheckpoint = &ch.checkpoints[index]
		currentCheckpoint := &ch.checkpoints[i]
		if i == 0 && height < currentCheckpoint.Height {
			return currentCheckpoint, i
		}
		if height < nextCheckpoint.Height && height >= currentCheckpoint.Height {
			return nextCheckpoint, i
		}
	}
```

internal/transports/p2p/peer/checkpoint.go
dorzepowski: in this solution it will return last checkpoint.
and actually it's not possible to reach this point if we can't find anything at this point,
because in first step we're looking if the checkpoint list is empty
and then we're checking if the height is equal or above last checkpoint from the list.
So it means that height should be between 0 and last checkpoint height, so for sure it should be there or am I missing something 🤔 ?

internal/transports/p2p/peer/checkpoint.go
dorzepowski: I'm not convinced to that
1. `checkpoint` seem to be good source for `checking` not verifying because it's not `verificationpoint` 😄 
2. I was thinking about being more clear method name, but it's hard decision because from the readability and abstraction point of view, `Check` seems more proper - it shouldn't bother you how it is implemented below, if it is storing the checkpoint or if it is searching for it each time. Maybe the only thing that should be kinda published is the error that we received an unexpected height that is above the last checkpoint 🤷 

To wrap up:
1. I'm rather voteing for `Check` instead of `Verify`, but if someone else would vote for verify, I can live with that (and rename it)
2. I'm not convinced if we really need to add `Update` or other suffix telling about side effects, so the third vote would be necessary
FYI @chris-4chain ☝️

internal/transports/p2p/peer/checkpoint.go
dorzepowski: My intention was to clarify which methods you're supposed to use and which methods are private to that component. I agree that it can be made private; however, since the checkpoint struct is private and is a private field, it effectively remains private to other packages. In other words, I'm treating this as a way to create more access scopes. If you find this confusing, I can make it private. From my perspective, though, making it private could confuse devs (writing the peer code) about which methods are available for use.

internal/transports/p2p/peer/checkpoint.go
dorzepowski: This is preparation for the case when we have multiple live connections but only one is syncing.
So imagine the case, when we're instantiating 8 peers, 
one is syncing and we're for example in the middle of the checkpoint list.
Then connection is dropped and we're switching to other peer. But he's checkpoint is outdated, so we need to actually set it at the point of starting sync. 
Maybe better solution would be to create this component only at the moment of StartingSync not at the moment of creation of the peer, like it is currently, WDYT?

internal/transports/p2p/peer/checkpoint_test.go
dorzepowski: I don't want to test method next, because it is intentionally private for this component, and it shouldn't bother us how it is working, so it's simpler to make any refactoring or splitting or anything.

internal/transports/p2p/peer/peer.go
dorzepowski: The possible solution in the comment above ☝️

internal/transports/p2p/peer/checkpoint.go
dorzepowski: The only situation when index would be out of range is when the currentIndex is the last index of that list,
this situation is already handled by the previous if statement, where we're checking if the height is equal or above last checkpoint, so I don't see a case, when it can be out of range, and actually I believe that I have a test case, which is checking the change from last checkpoint to the next one.

internal/transports/p2p/peer/checkpoint.go
```diff
@@ -0,0 +1,139 @@
...
+
+		ch.next(header.Height)
+		return nil
+	}
+
+	if header.Height > ch.checkpoint.Height {
+		return fmt.Errorf("unexpected header above next checkpoint height, got: %v, for checkpoint at height %d", header, ch.checkpoint.Height)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
dorzepowski: You're right 🤦

internal/transports/p2p/peer/checkpoint.go
chris-4chain: I you and me fell into this trap maybe this algo should be moved to another func or at least comment it why it is implemented like that.

internal/transports/p2p/peer/checkpoint.go
chris-4chain: > From my perspective, though, making it private could confuse devs (writing the peer code) about which methods are available for use.
Yes, I agree with you @dorzepowski. Besides, IMHO, it's crucial to point out that using `ch.checkpoint.Height` is not "thread-safe" - instead of the public Height method. 

Maybe we could add a comment somewhere about it.

internal/transports/p2p/peer/checkpoint.go
chris-4chain: > checkpoint seem to be good source for checking not verifying because it's not verificationpoint 😄

⬆️  It's funny but it is true 🙈 

About point `2.`. @kuba-4chain has a point that the name: `Check` suggests that it is a "const method".
> In C++ const methods `int function_name () const` are denied to change the state of the instance. 

In this case the `Check` methods modifies it...
The word "check" often implies a passive operation...
🤔 
🤔 
My types:
`CheckAndAdvance`, `Process`, `CheckAndProgress`, `Handle`

kuba-4chain: > checkpoint seem to be good source for checking not verifying because it's not verificationpoint 😄

> ⬆️ It's funny but it is true 🙈

But we're not exactly `checking` the `point`. What we're doing is `verifying` whether the checkpoint hash is equal to the block hash we received. We're actually `verifying` the block header against the checkpoint.

internal/transports/p2p/peer/checkpoint.go
kuba-4chain: But creating a new checkpoint at the moment of StartinSync is also I good idea IMO

internal/transports/p2p/peer/checkpoint_test.go
kuba-4chain: I think the method `findNextCheckpoint` is a method that should be private and should not bother us, but the `next()` method could be public. But it's just my opinion. Maybe @chris-4chain has an opinion about this as well? 😄

internal/transports/p2p/peer/checkpoint.go
dorzepowski: simplified

internal/transports/p2p/peer/checkpoint.go
dorzepowski: removed

internal/transports/p2p/peer/checkpoint.go
dorzepowski: method removed, and checkpoint is created at the start sync method

internal/transports/p2p/peer/checkpoint_test.go
dorzepowski: method removed

kuba-4chain (approved):

PR #242, opened 2024-04-15
feat(BUX-723): inbound peer logic implemented

kuba-4chain:
# Pull Request Checklist

### DIFF
--- a/internal/transports/p2p/peer/peer.go
+++ b/internal/transports/p2p/peer/peer.go
@@ -27,7 +27,7 @@ type Peer struct {
 	addr *net.TCPAddr
 	// listener is the net.Listener for inbound connections
 	listener net.Listener
-	// inbound is specifing if the peer is inbound (incoming) or outbound (outgoing)
+	// inbound is specifying if the peer is inbound (incoming) or outbound (outgoing)
 	inbound bool
 	// cfg is the P2P configuration specified by the user
 	cfg *config.P2PConfig
@@ -59,9 +59,9 @@ type Peer struct {
 	timeOffset int64
 	// userAgent is the 'name' of the peer used for identification
 	userAgent string
-	// syncedCheckpoints is specifing whether we have synced all checkpoints
+	// syncedCheckpoints is specifying whether we have synced all checkpoints
 	syncedCheckpoints bool
-	// sendHeadersMode is specifing whether we already sent the sendheaders message
+	// sendHeadersMode is specifying whether we already sent the sendheaders message
 	// to the peer and we're expecting to get just headers and no inv msg
 	sendHeadersMode bool
 
--- a/internal/transports/p2p/peer/peer.go
+++ b/internal/transports/p2p/peer/peer.go
@@ -139,7 +139,11 @@ func (p *Peer) Connect() error {
 	}
 
 	p.conn = conn
-	p.updatePeerAddr()
+	err = p.updatePeerAddr()
+	if err != nil {
+		_ = p.Disconnect()
+		return err
+	}
 
 	p.log.Info().Msgf("connected to peer: %s", p)
 
@@ -196,7 +200,6 @@ func (p *Peer) updatePeerAddr() error {
 	} else {
 		errMsg := "error retreiving address from peer"
 		p.log.Error().Msg(errMsg)
-		_ = p.Disconnect()
 		return errors.New(errMsg)
 	}
 	return nil
--- a/internal/transports/p2p/peer/peer.go
+++ b/internal/transports/p2p/peer/peer.go
@@ -23,6 +23,8 @@ import (
 type Peer struct {
 	conn              net.Conn
 	addr              *net.TCPAddr
+	listener          net.Listener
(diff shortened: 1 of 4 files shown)


internal/transports/p2p/server.go
```diff
@@ -36,11 +35,28 @@ func NewServer(
...
+	if err != nil {
+		return err
+	}
+
+	return s.listenAndConnect()
+}
+
+func (s *server) Shutdown() error {
+	for _, p := range s.peers {
+		_ = p.Disconnect()
```
chris-4chain, line 54: Maybe we should at least log the error instead of ignoring it 🤔 ?

internal/transports/p2p/server.go
```diff
@@ -56,9 +72,31 @@ func (s *server) Start() error {
...
+	peer, err := peer.NewInboundPeer(listener, s.config, s.chainParams, s.headersService, s.chainService, s.log)
+	if err != nil {
+		return err
+	}
+
+	return s.connectPeer(peer)
+}
+
+func (s *server) connectPeer(peer *peer.Peer) error {
+	s.peers = append(s.peers, peer)
```
chris-4chain, line 97: Is it designed to be executed from single goroutine or concurrently from many?

internal/transports/p2p/server.go
kuba-4chain: We're logging the error in `p.Disconnect()` method.

My idea for this implementation is to remove returning error from this method completely, and wait for Disconnection and/or error from all peers with WaitGroup in server shutdown. But this synchronisation will be done in the next task regarding multiple connections, while not returning the error can only be done after we decide to switch entirely to this solution and remove the interface for P2P that we temporarily created in the `main` method for this to work.

internal/transports/p2p/server.go
kuba-4chain: For now it's synchronous, but it should be rewritten to support concurrency in the next task regarding multiple connections.

But if you think I should support concurrency here already in this PR, let me know :)

internal/transports/p2p/peer/peer.go
```diff
@@ -75,6 +105,39 @@ func NewPeer(
...
+		quitting:          false,
+		quit:              make(chan struct{}),
+	}
+	return peer, nil
+}
+
+func NewInboundPeer(
+	listener net.Listener,
+	cfg *config.P2PConfig,
+	chainParams *chaincfg.Params,
```
dorzepowski, line 130: This is idea for another task, but maybe we should make chainParams a part of P2PConfig instead of retrieving it the way we currently do, WDYT?

internal/transports/p2p/peer/peer.go
```diff
@@ -94,12 +157,26 @@ func NewPeer(
...
 func (p *Peer) Connect() error {
-	conn, err := net.Dial(p.addr.Network(), p.addr.String())
+	var conn net.Conn
+	var err error
+
+	if p.inbound {
+		conn, err = p.listener.Accept()
+	} else {
+		conn, err = net.Dial(p.addr.Network(), p.addr.String())
+	}
```
dorzepowski, line 167: 🤔 Maybe it's a good idea to leave establishing connection to some higher component. 
In other words:
server (or any other component in the future) will establish the connection either outbound connection by Dialing or inbound connection by accepting, and pass this connection in a "constructor".
This way, inbound flag could be actually only informative for now, and later used for "connection management" to choose best outbound peers.
Additionally this will enable creation of unit tests with mocked connection instead of making it at the real network level.

internal/transports/p2p/server.go
dorzepowski: Actually this comment bring some additional thoughts to me,
I checked the code, and there is no place where we're actually handling the error from Disconnect method.
And I have some doubts:
1. How actually we should handle such an error, 
2. is log message enough? or we're able to do something more?
3. if log message (or internal peer things) are enough, then we shouldn't actually return anything (or at least we shouldn't return error type) because this is making the code only missleading.

internal/transports/p2p/server.go
dorzepowski: IMHO this is good moment to at least handle unlimited number of incoming connections. And just put adding some limits and managing list of those connections to the other task.
Because there are not so many code changes so I can predict that additional for loop couldn't make it hard to review, but I assume that adding this in a later stage, could make the later PR harder to read and actually it can be easily skipped because outbound connections have higher priority for us.

dorzepowski (changes requested):

internal/transports/p2p/server.go
kuba-4chain: Removed

internal/transports/p2p/peer/peer.go
kuba-4chain: You're right, I've created a task for that

chris-4chain (approved):

dorzepowski (approved):

chris-4chain:
testing 🧪

PR #246, opened 2024-04-26
fix(SPV-769): check sql.select result length

arkadiuszos4chain:
# Pull Request Checklist

### DIFF
--- a/service/header_service.go
+++ b/service/header_service.go
@@ -66,12 +66,12 @@ func (hs *HeaderService) IsCurrent() bool {
 	// Not current if the latest main (best) chain height is before the
 	// latest known good checkpoint (when checkpoints are enabled).
 	checkpoints := hs.checkpoints
-	checkpoint := &checkpoints[len(checkpoints)-1]
+	checkpoint := checkpoints[len(checkpoints)-1]
 	tip := hs.GetTip()
 	if tip == nil {
 		return true
 	}
-	if checkpoint != nil && tip.Height < checkpoint.Height {
+	if tip.Height < checkpoint.Height {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


service/header_service.go
```diff
@@ -66,12 +66,12 @@ func (hs *HeaderService) IsCurrent() bool {
 	// Not current if the latest main (best) chain height is before the
 	// latest known good checkpoint (when checkpoints are enabled).
 	checkpoints := hs.checkpoints
-	checkpoint := &checkpoints[len(checkpoints)-1]
+	checkpoint := checkpoints[len(checkpoints)-1]
```
chris-4chain, line 69: What if the `checkpoints` is empty?

chris-4chain (approved):

service/header_service.go
arkadiuszos4chain: I have no idea what are you talking about :P Change reverted - out of the scope

Nazarii-4chain (approved):

wregulski (approved):

kuba-4chain (commented):
In the method `GetAllTips()` in this file, we're also using `db.Select()`. If we're deciding on this approach of returning an error when we retrieve no records, then maybe we should add that in this `GetAllTips()` method as well?

jakubmkowalski:
testing

dorzepowski:
> In the method `GetAllTips()` in this file, we're also using `db.Select()`. If we're deciding on this approach of returning an error when we retrieve no records, then maybe we should add that in this `GetAllTips()` method as well?

Those changed methods where we're returning errors are semantically expected to return something, so when we don't have this something it is expected to get an error. 
While in case of GetAll**** it is expected to get all and if there isn't anything then get empty list. 
(It is a common design for such methods)

PR #247, opened 2024-04-26
feat(SPV-742): add peer manager

arkadiuszos4chain:
# Pull Request Checklist


cmd/main.go
```diff
@@ -42,7 +42,7 @@ var version = "development"
 
 type P2PServer interface {
 	Start() error
-	Shutdown() error
```
wregulski, line 45: Isn't it a good practice to stay with error return when shutting down? Probably we can't do anything with that, but at least we have a message and we can possibly kill it manually.

internal/chaincfg/params.go
```diff
@@ -108,7 +108,7 @@ type Params struct {
 	Net wire.BitcoinNet
 
 	// DefaultPort defines the default peer-to-peer port for the network.
-	DefaultPort string
+	DefaultPort uint16
```
wregulski, line 111: 👍

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/bitcoin-sv/block-headers-service/internal/transports/p2p/peer"
+	"github.com/bitcoin-sv/block-headers-service/internal/wire"
+)
+
+type AddressBook struct {
+	banDuration time.Duration
+	addrs       []*knownAddress
+	keyIndex    map[string]int
+
```
wregulski, line 17: Unnecessary line

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func NewAdressbook(banDuration time.Duration, acceptLocalAddresses bool) *AddressBook {
+	addrFitlerFn := IsRoutable
+	if acceptLocalAddresses {
+		addrFitlerFn = IsRoutableWithLocal
+	}
+
+	return &AddressBook{
+		addrs:        make([]*knownAddress, 0, 500),
```
wregulski, line 29: ### Minor 
I know that you won't like it, but maybe at least make 500 a const in the scope of this file? 🙈

internal/transports/p2p/server.go
```diff
@@ -29,99 +40,259 @@ func NewServer(
...
+
+			for ; peersToConnect > 0; peersToConnect-- {
+				s.connectToRandomAddr()
+			}
+		}
+	}
+}
+
+func (s *server) connectToRandomAddr() {
+	const tries = 20
```
wregulski, line 226: Maybe follow the same pattern as in Webhooks config, and add "MaxTries" to P2PConfig? 🤔

transports/p2p/connmgr/seed.go
```diff
@@ -60,15 +60,15 @@ func SeedFromDNS(chainParams *chaincfg.Params, reqServices wire.ServiceFlag,
 			}
 			addresses := make([]*wire.NetAddress, len(seedpeers))
 			// if this errors then we have *real* problems
-			intPort, _ := strconv.Atoi(chainParams.DefaultPort)
+			intPort := chainParams.DefaultPort //strconv.Atoi(chainParams.DefaultPort)
```
wregulski, line 63: Shouldn't we remove this comment?

cmd/main.go
arkadiuszos4chain: You can see that neither implementation returns any error - so... why stay here with an error?

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,128 @@
...
+		if ka.peer == nil {
+			if ka.isBanned(a.banDuration) {
+				continue
+			}
+			return ka.addr
+		}
+	}
+
+	return nil
+}
```
wregulski, line 96:
```
func (a *AddressBook) GetRndUnusedAddr(tries uint) *wire.NetAddress {
    a.mu.Lock()
    defer a.mu.Unlock()

    keyCount := len(a.keys)
    if keyCount == 0 {
        return nil
    }

    for i := uint(0); i < tries; i++ {
        // Randomly select an index and get the corresponding key
        rndIndex := rand.Intn(keyCount)
        ka := a.addrs[a.keys[rndIndex]]

        // Check if the address is unused and not banned
        if ka.peer == nil && !ka.isBanned(a.banDuration) {
            return ka.addr
        }
    }
    return nil
}

func (ka *knownAddress) isBanned(banDuration time.Duration) bool {
    return time.Now().Before(ka.bannedUntil)
}
```

transports/p2p/connmgr/seed.go
arkadiuszos4chain: Removed

internal/transports/p2p/network/address_book.go
arkadiuszos4chain: IMO it's unnecessary but moved to const

internal/transports/p2p/server.go
arkadiuszos4chain: It shouldn't be configurable because it's an implementation detail. 

In the end, we are attempting to establish and maintain a fixed number of connections in an infinite loop, so the value of this constant is not crucial. I use it simply to provide information to us (developers and administrators) that, over time, the implementation of the address book and its random address selection may no longer be sufficient.

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+			if ka.isBanned(a.banDuration) {
+				continue
+			}
+			return ka.addr
+		}
+	}
+
+	// return nil if no suitable address is found
+	return nil
+}
```
chris-4chain, line 106: Maybe the `tries` param is not necessary here. Consider this implementation:
It requires more memory (can be optimized), but we are sure that some address will be picked (if possible)
```golang
func (a *AddressBook) GetRndUnusedAddr() *wire.NetAddress {
	a.mu.Lock()
	defer a.mu.Unlock()

	free := make([]*KnownAddress, 0, len(a.addrs))
	for _, ka := range a.addrs {
		if ka.peer == nil && !ka.isBanned(a.banDuration) {
			free = append(free, ka)
		}
	}

	if len(free) == 0 {
		// return nil if no free address is found
		return nil
	}

	return free[rand.Intn(len(free))].addr
}
```

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
 	}
 
-	return s.listenAndConnect()
+	err = s.listenInboundPeers()
+	if err != nil {
+		s.log.Error().Msgf("error during server start. Shutdown p2p server. Reason: %v", err)
+		s.Shutdown()
+
+		return err
+	}
```
kuba-4chain, line 76: 1. Methods `connectOutboundPeers` and `listenInboundPeers` could be goroutines, but this can also be a next task to do for this package. WDYT?
2. We only shutdown when `listenInboundPeers` fails, but we don't shutdown when `connectOutboundPeers` fails? I think `server.Shutdown()` should happen in `main.go` when `server.Start()` returns an error. So I would remove the shutdown from here completely.

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
-func (s *server) seedAndConnect() error {
+func (s *server) connectOutboundPeers() error {
 	seeds := seedFromDNS(s.chainParams.DNSSeeds, s.log)
 	if len(seeds) == 0 {
 		return errors.New("no seeds found")
 	}
 
+	if len(seeds) > int(s.config.MaxOutboundConnections) {
+		seeds = seeds[:s.config.MaxOutboundConnections]
+	}
```
kuba-4chain, line 108: I think we should add the rest of the addresses from seeds to an AddressBook

kuba-4chain: Also, I'm wondering if we should also filter the seed addresses with our `IsRoutable` function from `network` 🤔 . But maybe we should just trust them, IDK

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
+		select {
+		case <-s.ctx.Done(): // Exit if context was canceled
+			s.log.Info().Msg("[observeInboundPeers] context canceled -> exit")
+			return
+		default:
+			s.ctxWg.Add(1) // Wait on shutdown
+
+			if s.inboundPeers.Space() == 0 {
+				s.log.Debug().Msg("[observeInboundPeers] nothing to do")
+				s.noWaitingSleep(sleeDuration)
```
kuba-4chain, line 257: We don't need to sleep here, because the operation `s.listener.Accept` in the `s.waitForIncomingConnection()` function is blocking.

internal/transports/p2p/peer/peers_collection.go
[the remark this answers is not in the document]
kuba-4chain: Also, based on the usage of function `Space()` i would change it to `HasSpace()` that returns a bool:

```go
// HasSpace returns true if there are available slots for new peers in the PeersCollection.
func (col *PeersCollection) HasSpace() bool {
	col.mu.Lock()
	defer col.mu.Unlock()

	return col.size - uint(len(col.peers)) > 0
}
```

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
+			peer.Disconnect()
+			return nil, err
+		}
+	}
+
+	return peer, nil
+}
+
+func (s *server) observeOutboundPeers() {
+	sleeDuration := 1 * time.Minute
```
kuba-4chain, line 206: Typo in `sleeDuration` :)

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
-		return err
+		peer.Disconnect()
+		s.log.Error().Str("peer", peer.String()).
+			Bool("inbound", inbound).
+			Msgf("error connecting with peer, reason: %v", err)
+		return nil, err
 	}
 
-	s.peers = append(s.peers, peer)
+	peer.SendGetAddrInfo()
```
kuba-4chain, line 192: I would put it as a part of `Connect()` function, after the write/read/ping handlers and make it private.

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
 
-	err = peer.Connect()
+	if !inbound {
+		err = peer.StartHeadersSync()
+		if err != nil {
+			peer.Disconnect()
+			return nil, err
+		}
+	}
```
kuba-4chain, line 200: This part cannot be here, because it will start a sync with every peer we connect to. And we wanted to just sync from one peer.

I suggest, in the scope of this task, to start sync with the first connected peer, or with the first from the list or a random peer. And we have a [separate task SPV-744](https://bitcoinassociation.atlassian.net/browse/SPV-744) for choosing the best sync peer.

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
+			if s.inboundPeers.Space() == 0 {
+				s.log.Debug().Msg("[observeInboundPeers] nothing to do")
+				s.noWaitingSleep(sleeDuration)
+				continue
+			}
+
+			s.log.Info().Msgf("listening for inbound connections on port %d", s.chainParams.DefaultPort)
+			s.waitForIncomingConnection() // Accept connection one-by-one to gracefully handle shutdown
+
+			s.ctxWg.Done()
```
dorzepowski, line 264: It looks a little complicated. 
IMHO we should run in go routine an infinite loop that 
will accept connections, 
try to add it to PeersCollection (or check if there is a space)
if it is ok to add it, then use peer.Connect
if it is not ok, then disconnect the connection


And in case of exiting, all we need to do is: `s.listener.Close()`

dorzepowski (changes requested):

internal/transports/p2p/server.go
chris-4chain: SleeDuration sounds cool 😎

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/bitcoin-sv/block-headers-service/internal/transports/p2p/peer"
+	"github.com/bitcoin-sv/block-headers-service/internal/wire"
+)
+
+// AddressBook represents a collection of known network addresses.
+type AddressBook struct {
+	banDuration  time.Duration
+	addrs        []*knownAddress // addrs is a slice containing known addresses
+	addrsLookup  map[string]int  // addrLookup is a map for fast lookup of addresses, maps address key to index in addrs slice
```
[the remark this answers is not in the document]
arkadiuszos4chain: We will have ~2k addresses. I'm not sure that allocating new array of this size every time we want random address is a good idea.

internal/transports/p2p/network/network.go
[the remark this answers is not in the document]
arkadiuszos4chain: IMO it's a package of helper function so I would leave it here

internal/transports/p2p/server.go
arkadiuszos4chain: fixed

internal/transports/p2p/network/address_book_test.go
```diff
@@ -0,0 +1,129 @@
...
+		// when
+		sut.BanAddr(addr)
+
+		// then
+		require.True(t, sut.addrs[0].isBanned(time.Hour))
+	})
+}
+
+func TestAddressBook_GetRndUnusedAddr(t *testing.T) {
+	t.Run("", func(t *testing.T) {
```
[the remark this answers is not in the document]
arkadiuszos4chain: fixed

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+
+	_, ka := a.internalFind(addr)
+	if ka != nil {
+		now := time.Now()
+		ka.banTimestamp = &now
+	}
+}
+
+// GetRndUnusedAddr returns a randomly chosen unused network address.
+func (a *AddressBook) GetRndUnusedAddr(tries uint) *wire.NetAddress {
```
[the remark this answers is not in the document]
arkadiuszos4chain: fixed

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+type AddressBook struct {
+	banDuration  time.Duration
+	addrs        []*knownAddress // addrs is a slice containing known addresses
+	addrsLookup  map[string]int  // addrLookup is a map for fast lookup of addresses, maps address key to index in addrs slice
+	mu           sync.Mutex
+	addrFitlerFn func(*wire.NetAddress) bool
+}
+
+// NewAddressBook creates and initializes a new AddressBook instance.
+func NewAdressBook(banDuration time.Duration, acceptLocalAddresses bool) *AddressBook {
```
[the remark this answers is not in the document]
arkadiuszos4chain: fixed

config/defaults.go
```diff
@@ -72,6 +72,8 @@ func getP2PDefaults() *P2PConfig {
 		UserAgentName:             ApplicationName,
 		UserAgentVersion:          Version(),
 		Experimental:              false,
+		MaxOutboundConnections:    8,
+		MaxInboundConnections:     8,
```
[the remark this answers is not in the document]
arkadiuszos4chain: removed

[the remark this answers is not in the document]
arkadiuszos4chain: Done

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+			return ka.addr
+		}
+	}
+
+	// return nil if no suitable address is found
+	return nil
+}
+
+func (a *AddressBook) internalFind(addr *wire.NetAddress) (string, *knownAddress) {
```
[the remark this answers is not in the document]
arkadiuszos4chain: changed to `findAddr` `addAddr`

internal/transports/p2p/peer/peer.go
```diff
@@ -19,11 +19,18 @@ import (
...
+	SignalError(*Peer, error)
+}
+
 type Peer struct {
 	// conn is the current connection to peer
 	conn net.Conn
 	// addr is the net.Addr of peer, used for connection
-	addr *net.TCPAddr
+	addr  *net.TCPAddr
+	addrS string
```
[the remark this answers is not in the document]
arkadiuszos4chain: To avoid multpiple string concatecation every time we log something. I need to check if still need this with new logger instance

internal/transports/p2p/peer/peer.go
```diff
@@ -591,5 +629,21 @@ func (p *Peer) isSynced() bool {
...
+func (p *Peer) logInfo(format string, v ...any) {
+	p.log.Info().Str("peer", p.String()).Bool("inbound", p.inbound).Msgf(format, v...)
+}
+
+func (p *Peer) logWrn(format string, v ...any) {
+	p.log.Warn().Str("peer", p.String()).Bool("inbound", p.inbound).Msgf(format, v...)
+}
+
+func (p *Peer) logError(format string, v ...any) {
+	p.log.Error().Str("peer", p.String()).Bool("inbound", p.inbound).Msgf(format, v...)
```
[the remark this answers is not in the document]
arkadiuszos4chain: What an elegant idea. Changed

internal/transports/p2p/peer/peer.go
```diff
@@ -591,5 +629,21 @@ func (p *Peer) isSynced() bool {
...
+
+func (p *Peer) logDebug(format string, v ...any) {
+	p.log.Debug().Str("peer", p.String()).Bool("inbound", p.inbound).Msgf(format, v...)
+}
+
+func (p *Peer) logInfo(format string, v ...any) {
+	p.log.Info().Str("peer", p.String()).Bool("inbound", p.inbound).Msgf(format, v...)
+}
+
+func (p *Peer) logWrn(format string, v ...any) {
```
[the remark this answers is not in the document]
arkadiuszos4chain: Replaced with new logger instance

internal/transports/p2p/server.go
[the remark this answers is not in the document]
arkadiuszos4chain: fixed

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+	addrs        []*knownAddress // addrs is a slice containing known addresses
+	addrsLookup  map[string]int  // addrLookup is a map for fast lookup of addresses, maps address key to index in addrs slice
+	mu           sync.Mutex
+	addrFitlerFn func(*wire.NetAddress) bool
+}
+
+// NewAddressBook creates and initializes a new AddressBook instance.
+func NewAdressBook(banDuration time.Duration, acceptLocalAddresses bool) *AddressBook {
+	// Set the address filter function based on whether local addresses are accepted
+	addrFitlerFn := IsRoutable
```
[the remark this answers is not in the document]
arkadiuszos4chain: fixed

internal/transports/p2p/server.go
arkadiuszos4chain: What is wrong with syncing with multiple nodes?

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+
+type knownAddress struct {
+	addr *wire.NetAddress
+	peer *peer.Peer
+
+	banTimestamp *time.Time
+}
+
+func (a *knownAddress) isBanned(duration time.Duration) bool {
+	return a.banTimestamp != nil && time.Since(*a.banTimestamp) < duration
```
kuba-4chain, line 137: There's one more problem here. Once the peer is banned, we never unban him. Meaning, we never make `banTimestampt` to nil. So once the address is banned, even if the duration passes, we will still return that it's banned.

We probably need some handler for unbanning peers, either periodically every idk, 5-10 minutes, or during some operations, maybe even in this function as a part of checking if the address in banned.

internal/transports/p2p/server.go
kuba-4chain: 1. Unnecessary network load when syncing from 0. While syncing from 0, we ask for very old headers, which every node should have. So there's no need for asking multiple nodes to send us old headers, one is enough. Other nodes should be focused on keeping the network up to date with newest headers. Remember that when syncing we ask for batches of 2 000 headers every second or so. That's not a super small network load.
2. Unnecessary network load when synced. When synced and waiting for new headers, we could ask multiple nodes to send us the headers, so we can time the response and give a score to each peer based on their response time and then sync with just that one. But that's for another task IMHO.
3. DB load and synchronisation problems. When asking multiple peers for old nodes in batches, multiple peers are gonna send us 2 000 headers in more or less the same time every second. Then we have to put them into DB, and after the first batch is inserted all other batches are gonna return an error "headers already in db". And if we start blocking the db for synchronisation, we'll have race conditions and slow down the entire sync process significantly. Why would we do this to ourselves? 😬

internal/transports/p2p/peer/peers_collection.go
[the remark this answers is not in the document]
arkadiuszos4chain: > I'm not really convinced about the whole `peers_collection` implementation using an array of slots, where every method has linear complexity. Consider something like this:
> 
> ```go
> package transactions
> 
> import (
> 	"errors"
> 	"slices"
> 	"sync"
> )
> 
> // PeersCollection represents a fixed size collection of peer objects with concurrency-safe operations.
> type PeersCollection struct {
> 	peers []*Peer
> 	size  uint
> 	mu    sync.Mutex
> }
> 
> // NewPeersCollection creates and initializes a new PeersCollection instance with the specified, fixed size.
> func NewPeersCollection(size uint) *PeersCollection {
> 	return &PeersCollection{
> 		size:  size,
> 		peers: make([]*Peer, 0, size),
> 	}
> }
> 
> // AddPeer adds a new peer to the PeersCollection.
> // Returns an error if there is no space available for the new peer.
> func (col *PeersCollection) AddPeer(p *Peer) error {
> 	col.mu.Lock()
> 	defer col.mu.Unlock()
> 
> 	if len(col.peers) < int(col.size) {
> 		return errors.New("no space available for new peer")
> 	}
> 
> 	col.peers = append(col.peers, p)
> 	return nil
> }
> 
> // RmPeer removes the specified peer from the PeersCollection. Ignores address if doesn't exist in the PeersCollection.
> func (col *PeersCollection) RmPeer(p *Peer) {
> 	col.mu.Lock()
> 	defer col.mu.Unlock()
> 
> 	// find index of peer
> 	pIndex := slices.Index(col.peers, p)
> 	if pIndex == -1 {
> 		return // could return an error here
> 	}
> 
> 	// substitute with last element
> 	col.peers[pIndex] = col.peers[len(col.peers)-1]
> 
> 	// remove last element
> 	col.peers = col.peers[:len(col.peers)-1]
> }
> 
> // Space returns the number of available slots for new peers in the PeersCollection.
> func (col *PeersCollection) Space() uint {
> 	col.mu.Lock()
> 	defer col.mu.Unlock()
> 
> 	return col.size - uint(len(col.peers))
> }
> 
> // Enumerate returns a slice containing all non-nil peers in the PeersCollection. Order of peers in the returned slice is not guaranteed.
> func (col *PeersCollection) Enumerate() []*Peer {
> 	col.mu.Lock()
> 	defer col.mu.Unlock()
> 
> 	// copy slice
> 	res := make([]*Peer, len(col.peers))
> 	copy(res, col.peers)
> 
> 	return res
> }
> ```

It ismuch better than mine

internal/transports/p2p/network/address_book.go
arkadiuszos4chain: 1. Firstly, it's a naive and straightforward ban implementation by design.
2. We don't need to "unban" an address because if the timestamp is older than the ban duration, the address is automatically no longer banned.

internal/transports/p2p/server.go
```diff
@@ -29,99 +43,271 @@ func NewServer(
...
+		if err := s.connectToAddr(addr, s.chainParams.DefaultPort); err != nil {
+			continue
+		}
+
+		peersCounter++
+	}
+
+	if peersCounter == 0 {
+		return errors.New("cannot connect to any peers from seed")
 	}
```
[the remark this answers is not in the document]
arkadiuszos4chain: Make sense

internal/transports/p2p/network/address_book.go
kuba-4chain: Okay, you're right, I'm stupid and misunderstood the return statement

internal/transports/p2p/network/network.go
```diff
@@ -0,0 +1,291 @@
...
+
+// ipNet returns a net.IPNet struct given the passed IP address string, number
+// of one bits to include at the start of the mask, and the total number of bits
+// for the mask.
+func ipNet(ip string, ones, bits int) net.IPNet {
+	return net.IPNet{IP: net.ParseIP(ip), Mask: net.CIDRMask(ones, bits)}
+}
+
+// IsIPv4 returns whether or not the given address is an IPv4 address.
+func IsIPv4(na *wire.NetAddress) bool {
```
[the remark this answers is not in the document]
arkadiuszos4chain: Moved to wire package as @kuba-4chain suggested

internal/transports/p2p/network/address_book.go
arkadiuszos4chain: All ban logic has been updated (along with the address book). Now, when we ban an address, it is moved to the banned collection, and a new goroutine is started. This goroutine is responsible for moving the address back to the free (unbanned, ready to use) collection when the ban is over

internal/transports/p2p/server.go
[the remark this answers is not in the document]
arkadiuszos4chain: Changed to connect to one randomly chosen address

internal/transports/p2p/network/address_book.go
```diff
@@ -0,0 +1,138 @@
...
+}
```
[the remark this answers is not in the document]
arkadiuszos4chain: Address book was reimplemented

internal/transports/p2p/server.go
arkadiuszos4chain: Good point. Now they are stored in address book and filtered with IsRoutable by the book

internal/transports/p2p/server.go
arkadiuszos4chain: The 'tries' was removed. The AddressBook implementation was changed, and it is no longer required.

internal/transports/p2p/server.go
arkadiuszos4chain: Waiting (at least for now) is necessary here because we want to wait for some time before the next check to see if we have a free slot for an incoming connection.

internal/transports/p2p/server.go
arkadiuszos4chain: Clever, changed. But, to be honest, I'm not sure if it's as error-proof as the previous implementation

internal/transports/p2p/server.go
arkadiuszos4chain: implementation changed

internal/transports/p2p/server.go
arkadiuszos4chain: Changed. We use only one peer for sync from 0 now.

PR #252, opened 2024-05-14
feat(SPV-748): add response to getHeaders

jakubmkowalski:
# Pull Request Checklist

Adds responding to getHeaders message in p2p service

### DIFF
--- a/database/repository/header_repository.go
+++ b/database/repository/header_repository.go
@@ -186,10 +186,10 @@ func (r *HeaderRepository) GetHeadersByHeightRange(from int, to int) ([]*domains
 	return dto.ConvertToBlockHeader(bh), nil
 }
 
-func (r *HeaderRepository) GetHashStopHeight(hashStop string) (int32, error) {
-	hsh, err := r.db.GetHashStopHeight(hashStop)
+func (r *HeaderRepository) GetHeadersStopHeight(hashStop string) (int, error) {
+	hs, err := r.db.GetHeadersStopHeight(hashStop)
 	if err != nil {
 		return 0, err
 	}
-	return hsh, nil
+	return hs, nil
 }
--- a/database/repository/header_repository.go
+++ b/database/repository/header_repository.go
@@ -169,3 +169,27 @@ func (r *HeaderRepository) GetChainBetweenTwoHashes(low string, high string) ([]
 	}
 	return nil, err
 }
+
+func (r *HeaderRepository) GetHeadersHeightOfLocators(hashtable []interface{}, hashStop *chainhash.Hash) ([]*domains.BlockHeader, error) {
+	bh, err := r.db.GetHeadersHeightOfLocators(hashtable, hashStop)
+	if err != nil {
+		return nil, err
+	}
+	return dto.ConvertToBlockHeader(bh), nil
+}
+
+func (r *HeaderRepository) GetHeadersBetweenHeights(from int, to int) ([]*domains.BlockHeader, error) {
+	bh, err := r.db.GetHeadersBetweenHeights(from, to)
+	if err != nil {
+		return nil, err
+	}
+	return dto.ConvertToBlockHeader(bh), nil
+}
+
+func (r *HeaderRepository) GetHashStopHeight(hashStop string) (int32, error) {
+	hsh, err := r.db.GetHashStopHeight(hashStop)
+	if err != nil {
+		return 0, err
+	}
+	return hsh, nil
+}
--- a/database/repository/header_repository.go
+++ b/database/repository/header_repository.go
@@ -170,6 +170,7 @@ func (r *HeaderRepository) GetChainBetweenTwoHashes(low string, high string) ([]
 	return nil, err
 }
 
+// GetHeadersStartHeight returns height of the highest header from the list of hashes.
 func (r *HeaderRepository) GetHeadersStartHeight(hashtable []string) (int, error) {
 	sh, err := r.db.GetHeadersStartHeight(hashtable)
 	if err != nil {
@@ -178,6 +179,7 @@ func (r *HeaderRepository) GetHeadersStartHeight(hashtable []string) (int, error
 	return sh, nil
 }
 
+// GetHeadersByHeightRange returns headers from db in specified height range.
 func (r *HeaderRepository) GetHeadersByHeightRange(from int, to int) ([]*domains.BlockHeader, error) {
 	bh, err := r.db.GetHeadersByHeightRange(from, to)
 	if err != nil {
@@ -186,6 +188,7 @@ func (r *HeaderRepository) GetHeadersByHeightRange(from int, to int) ([]*domains
 	return dto.ConvertToBlockHeader(bh), nil
 }
 
+// GetHeadersStopHeight returns height of hashstop header from db.
 func (r *HeaderRepository) GetHeadersStopHeight(hashStop string) (int, error) {
 	hs, err := r.db.GetHeadersStopHeight(hashStop)
 	if err != nil {
--- a/database/repository/header_repository.go
+++ b/database/repository/header_repository.go
@@ -178,8 +178,8 @@ func (r *HeaderRepository) GetHeadersHeightOfLocators(hashtable []interface{}, h
 	return dto.ConvertToBlockHeader(bh), nil
 }
 
-func (r *HeaderRepository) GetHeadersBetweenHeights(from int, to int) ([]*domains.BlockHeader, error) {
-	bh, err := r.db.GetHeadersBetweenHeights(from, to)
+func (r *HeaderRepository) GetHeadersByHeightRange(from int, to int) ([]*domains.BlockHeader, error) {
+	bh, err := r.db.GetHeadersByHeightRange(from, to)
 	if err != nil {
 		return nil, err
 	}
--- a/database/repository/header_repository.go
+++ b/database/repository/header_repository.go
@@ -170,12 +170,12 @@ func (r *HeaderRepository) GetChainBetweenTwoHashes(low string, high string) ([]
 	return nil, err
 }
 
-func (r *HeaderRepository) GetHeadersHeightOfLocators(hashtable []interface{}, hashStop *chainhash.Hash) ([]*domains.BlockHeader, error) {
-	bh, err := r.db.GetHeadersHeightOfLocators(hashtable, hashStop)
+func (r *HeaderRepository) GetHeadersStartHeight(hashtable []string) (int, error) {
+	sh, err := r.db.GetHeadersStartHeight(hashtable)
 	if err != nil {
-		return nil, err
+		return 0, err
 	}
-	return dto.ConvertToBlockHeader(bh), nil
+	return sh, nil
 }
 
 func (r *HeaderRepository) GetHeadersByHeightRange(from int, to int) ([]*domains.BlockHeader, error) {
--- a/database/sql/headers.go
+++ b/database/sql/headers.go
@@ -10,6 +10,8 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/bitcoin-sv/block-headers-service/domains"
+	"github.com/bitcoin-sv/block-headers-service/internal/chaincfg/chainhash"
+	"github.com/bitcoin-sv/block-headers-service/internal/wire"
 	"github.com/bitcoin-sv/block-headers-service/repository/dto"
 )
 
@@ -160,6 +162,31 @@ const (
 	sqlTipOfChainHeight = `SELECT MAX(height) FROM headers WHERE header_state = 'LONGEST_CHAIN'`
 
 	sqlVerifyHash = `SELECT hash FROM headers WHERE merkleroot = $1 AND height = $2 AND header_state = 'LONGEST_CHAIN'`
+
+	sqlGetHeadersBeginning = `
+	WITH AllHashes (hash) AS (
+		VALUES 
+		(?)
+	`
+
+	sqlGetHeadersHash = `,
+		(?)`
+
+	sqlGetHeadersEnd = `
+	)
+	SELECT 
+    	COALESCE(h.height, '0') AS height,
+    	a.hash
+	FROM AllHashes a
+	LEFT JOIN headers h ON a.hash = h.hash;
+	`
+
+	sqlHeaderByHeightRangeLongestChain = `
+	SELECT 
+		hash, height, version, merkleroot, nonce, bits, chainwork, previous_block, timestamp, header_state, cumulated_work
+	FROM headers
+	WHERE height BETWEEN ? AND ? AND header_state = 'LONGEST_CHAIN';
+	`
 )
 
 // HeadersDb represents a database connection and map of related sql queries.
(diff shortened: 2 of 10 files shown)


database/sql/headers.go
```diff
@@ -396,6 +423,42 @@ func (h *HeadersDb) GetMerkleRootsConfirmations(
 	return confirmations, nil
 }
 
+// GetHeadersHeightOfLocators returns hash and height from db with given locators.
+func (h *HeadersDb) GetHeadersHeightOfLocators(hashtable []interface{}, hashStop *chainhash.Hash) (bh []*dto.DbBlockHeader, err error) {
```
wregulski, line 427: What do you think about splitting this func to two parts?

It may make the GetHeadersHeightOfLocators method cleaner and more focused on executing the query and handling the result.

```go
func BuildHashesQuery(hashtable []interface{}) string {
	query := sqlGetHeadersBeginning
	for i := 1; i < len(hashtable); i++ {
		query += sqlGetHeadersHash
	}
	return query + sqlGetHeadersEnd
}

func (h *HeadersDb) GetHeadersHeightOfLocators(hashtable []interface{}, hashStop *chainhash.Hash) ([]*dto.DbBlockHeader, error) {
	query := BuildHashesQuery(hashtable)
	var headers []*dto.DbBlockHeader
	if err := h.db.Select(&headers, h.db.Rebind(query), hashtable...); err != nil {
		h.log.Error().Err(err).Msg("Failed to get headers by locators")
		return nil, err
	}
	return headers, nil
}
```

database/sql/headers.go
```diff
@@ -396,6 +423,42 @@ func (h *HeadersDb) GetMerkleRootsConfirmations(
...
+
+	if err := h.db.Select(&bh, h.db.Rebind(query), hashtable...); err != nil {
+		h.log.Error().Err(err).Msg("Failed to get headers by height range")
+		return nil, err
+	}
+	return bh, nil
+}
+
+// GetHashStopHeight will return header from db with given hash.
+func (h *HeadersDb) GetHashStopHeight(hashStop string) (int32, error) {
```
wregulski, line 442: A bit simplified:

```go
func (h *HeadersDb) GetHashStopHeight(hashStop string) (int32, error) {
	var dbHashStop dto.DbBlockHeader
	err := h.db.Get(&dbHashStop, h.db.Rebind(sqlHeader), hashStop)
	if err != nil {
		if errors.Is(err, sql.ErrNoRows) {
			return wire.MaxCFHeadersPerMsg, nil
		}
		return 0, errors.Wrapf(err, "failed to get stophash %s", hashStop)
	}
	return dbHashStop.Height, nil
}
```

internal/transports/p2p/peer/peer.go
```diff
@@ -555,6 +557,27 @@ func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
 	}
 }
 
+func (p *Peer) handleGetHeadersMsg(msg *wire.MsgGetHeaders) {
+
```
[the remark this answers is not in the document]
jakubmkowalski: removed

database/repository/header_repository.go
```diff
@@ -169,3 +169,27 @@ func (r *HeaderRepository) GetChainBetweenTwoHashes(low string, high string) ([]
...
+
+func (r *HeaderRepository) GetHeadersHeightOfLocators(hashtable []interface{}, hashStop *chainhash.Hash) ([]*domains.BlockHeader, error) {
+	bh, err := r.db.GetHeadersHeightOfLocators(hashtable, hashStop)
+	if err != nil {
+		return nil, err
+	}
+	return dto.ConvertToBlockHeader(bh), nil
+}
+
+func (r *HeaderRepository) GetHeadersBetweenHeights(from int, to int) ([]*domains.BlockHeader, error) {
```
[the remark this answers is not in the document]
jakubmkowalski: changed

database/sql/headers.go
jakubmkowalski: query construction moved to new function

database/sql/headers.go
jakubmkowalski: function simplified

Nazarii-4chain (approved):

database/sql/headers.go
```diff
@@ -34,6 +36,12 @@ const (
 	WHERE hash = ?
 	`
 
+	sqlHeaderFromHashAndState = `
+	SELECT hash, height, version, merkleroot, nonce, bits, chainwork, previous_block, timestamp, header_state, cumulated_work
```
dorzepowski, line 40: Are we using it somewhere else then for stop height,
if not, then I would change this to select only height. This way, it would probably use only an index without reaching for actual data.

internal/tests/testrepository/header_testrepository.go
```diff
@@ -238,6 +238,18 @@ func (r *HeaderTestRepository) GetMerkleRootsConfirmations(
...
+}
+
+func (r *HeaderTestRepository) GetHeadersByHeightRange(from int, to int) ([]*domains.BlockHeader, error) {
+	return nil, nil
+}
+
+func (r *HeaderTestRepository) GetHeadersStopHeight(hashStop string) (int, error) {
+	return 0, nil
+}
+
```
dorzepowski, line 252: maybe it's better to panic here with message "not implemented"

internal/transports/p2p/peer/peer.go
```diff
@@ -555,6 +557,26 @@ func (p *Peer) handleHeadersMsg(msg *wire.MsgHeaders) {
...
+func (p *Peer) handleGetHeadersMsg(msg *wire.MsgGetHeaders) {
+	p.log.Info().Msgf("received getheaders msg from peer %s", p)
+	if !p.syncedCheckpoints {
+		p.log.Info().Msgf("we are still syncing, ignoring getHeaders msg from peer %s", p)
+		return
+	}
+
+	locator := msg.BlockLocatorHashes
+	if len(locator) == 0 {
+		p.log.Info().Msgf("no locator hashes in getheaders msg from peer %s", p)
```
dorzepowski, line 569: maybe it's better to log.Warn

service/header_service.go
```diff
@@ -291,6 +292,66 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
...
+	}
+
+	// Check if peer requested number of headers is higher than the maximum number of headers per message
+	if wire.MaxCFHeadersPerMsg < stopHeight-startHeight {
+		stopHeight = startHeight + wire.MaxCFHeadersPerMsg
+	}
+
+	dbHeaders, err := hs.repo.Headers.GetHeadersByHeightRange(startHeight+1, stopHeight)
+	if err != nil {
+		log.Trace().Msgf("Error getting headers between heights: %v", err)
```
dorzepowski, line 335: maybe it's good idea to log it as Error, because this error will rather suggest that we have some problem with db.

service/header_service.go
```diff
@@ -291,6 +292,66 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
...
+	hashes := make([]string, len(locators))
+	for i, v := range locators {
+		hashes[i] = v.String()
+	}
+
+	startHeight, err := hs.repo.Headers.GetHeadersStartHeight(hashes)
+	if err != nil {
+		log.Trace().Msgf("Error getting headers of locators: %v", err)
+		return nil
+	}
```
dorzepowski, line 311: Hmmm... maybe it's good idea actually to return errors from HeadersService, and log.Error at the Peer side, so it could add some context of the peer to the message (like IP, Client Agent, etc.)

service/header_service.go
```diff
@@ -291,6 +292,66 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
...
+			Bits:       dbHeader.Bits,
+			Nonce:      dbHeader.Nonce,
+		}
+		headers = append(headers, header)
+	}
+
+	return headers
+}
+
 // LocateHeaders fetches headers for a number of blocks after the most recent known block
```
dorzepowski, line 355: I wonder if we could replace this method with the new one,
or eventually call from that method a new method with translated params, so the private locateHeaders could be removed.

service/header_service.go
```diff
@@ -291,6 +292,66 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
 	return hs.repo.Headers.GetMerkleRootsConfirmations(request, hs.merkleCfg.MaxBlockHeightExcess)
 }
 
+// LocateHeadersGetHeaders returns headers with given hashes.
+func (hs *HeaderService) LocateHeadersGetHeaders(locators []*chainhash.Hash, hashstop *chainhash.Hash) []*wire.BlockHeader {
```
dorzepowski, line 296: Overall very good job with this implementation.

The one thing that is missing and is rather hard challange is to add tests for this method.

dorzepowski (changes requested):

database/sql/headers.go
jakubmkowalski: now we only take height

internal/tests/testrepository/header_testrepository.go
jakubmkowalski: implementation added

service/header_service.go
jakubmkowalski: changed to handle errors log at the peer side

service/header_service.go
jakubmkowalski: changed to handle errors log at the peer side

internal/transports/p2p/peer/peer.go
jakubmkowalski: changed to handle errors log at the peer side

service/header_service.go
jakubmkowalski: locateHeaders removed, now LocateHeaders uses locateHeadersGetHeaders function

service/header_service.go
jakubmkowalski: tests added

internal/tests/testrepository/header_testrepository.go
```diff
@@ -238,6 +238,39 @@ func (r *HeaderTestRepository) GetMerkleRootsConfirmations(
...
+func (r *HeaderTestRepository) GetHeadersStartHeight(hashtable []string) (int, error) {
+	for i := len(*r.db) - 1; i >= 0; i-- {
+		header := (*r.db)[i]
+		for j := len(hashtable) - 1; j >= 0; j-- {
+			if header.Hash.String() == hashtable[j] && header.State == domains.LongestChain {
+				return int(header.Height), nil
+			}
+		}
+	}
+	return 0, nil
```
dorzepowski, line 250: Just as a reminder, if you change a logic in a real sql, you should change it here also.

service/header_service_test.go
```diff
@@ -348,6 +349,153 @@ func TestMerkleRootConfirmations(t *testing.T) {
...
+		name                 string
+		locator              []*chainhash.Hash
+		hashstop             *chainhash.Hash
+		result               []*wire.BlockHeader
+		expectedErrorMessage string
+	}{
+		{
+			name: "happy path, all block locator hashes are valid (hashStart height equals 3), hashstop is valid (height equals 4), returns block with height 4",
+			locator: []*chainhash.Hash{
+				toChainhashPtr("0000000082b5015589a3fdf2d4baff403e6f0be035a5d9742c1cae6295464449"),
```
dorzepowski, line 365: there is an util method to create hash from string:

```suggestion
			fixtures.HashOf("0000000082b5015589a3fdf2d4baff403e6f0be035a5d9742c1cae6295464449"),
```

maybe it's good idea to use it.
Unfortunately it doesn't handle errors inside, so it would be good if you would add panic when error is there.

service/header_service_test.go
```diff
@@ -348,6 +349,153 @@ func TestMerkleRootConfirmations(t *testing.T) {
...
+				assert.Equal(t, err.Error(), tc.expectedErrorMessage)
+			} else {
+				for j, r := range result {
+					assert.Equal(t, r.Version, tc.result[j].Version)
+					assert.Equal(t, r.PrevBlock, tc.result[j].PrevBlock)
+					assert.Equal(t, r.MerkleRoot, tc.result[j].MerkleRoot)
+					assert.Equal(t, r.Bits, tc.result[j].Bits)
+					assert.Equal(t, r.Nonce, tc.result[j].Nonce)
+				}
+			}
```
dorzepowski, line 494: I know that it's tempting to put single test logic and drive it with test cases (I must admit that often I'm falling into this "trap")

But for me it is much clear when you have happy path tests cases done in one test and errors in the second.
Another argument for that is the fact that your introducing a branching in test code, so it can have easily a bug, and we don't have (and don't want to have) tests for tests to ensure that it is working as expected especially in the if statement.

service/header_service_test.go
```diff
@@ -385,3 +533,21 @@ func createHeader(height int32, hash chainhash.Hash, prevBlock chainhash.Hash) d
 		Chainwork:     big.NewInt(4295032833),
 	}
 }
+
+func toChainhash(hash string) chainhash.Hash {
```
dorzepowski, line 537: maybe it could be moved next to `fixutres.HashOf` function and named as `HashValueOf`
or you can simply use the HashOf method in the following way: `*fixtures.HashOf("....")` to get a value instead of pointer.

dorzepowski (changes requested):

database/sql/headers.go
```diff
@@ -160,6 +168,20 @@ const (
 	sqlTipOfChainHeight = `SELECT MAX(height) FROM headers WHERE header_state = 'LONGEST_CHAIN'`
 
 	sqlVerifyHash = `SELECT hash FROM headers WHERE merkleroot = $1 AND height = $2 AND header_state = 'LONGEST_CHAIN'`
+
+	sqlGetHeadersHeight = `
+	SELECT COALESCE(MAX(height), 0) AS startHeight
```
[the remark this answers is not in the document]
jakubmkowalski: According to the documentation
https://wiki.bitcoinsv.io/index.php/Peer-To-Peer_Protocol#getblocks
`Note that it is allowed to send in fewer known hashes down to a minimum of just one hash. However, the purpose of the block locator object is to detect a wrong branch in the caller's main chain. If the peer detects that you are off the main chain, it will send in block hashes which are earlier than your last known block. So if you just send in your last known hash and it is off the main chain, the peer starts over at block #1.`
and
`For the block locator object in this packet, the same rules apply as for the getblocks packet.`

service/header_service.go
```diff
@@ -291,86 +291,84 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
...
 	}
 
-	return startNode, total
+	return headers, nil
+}
+
+// LocateHeaders fetches headers for a number of blocks after the most recent known block
+// in the best chain, based on the provided block locator and stop hash, and defaults to the
+// genesis block if the locator is unknown.
+func (hs *HeaderService) LocateHeaders(locator domains.BlockLocator, hashStop *chainhash.Hash) ([]*wire.BlockHeader, error) {
```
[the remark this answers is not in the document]
jakubmkowalski: The original LocateHeaders signature remains unchanged, changes in serverpeer reverted

service/header_service.go
```diff
@@ -291,86 +291,84 @@ func (hs *HeaderService) GetMerkleRootsConfirmations(
...
+	if stopHeight == 0 {
+		stopHeight = startHeight + wire.MaxCFHeadersPerMsg
 	}
-	startNode = next
 
-	// Calculate how many entries are needed.
-	total := uint32((hs.GetTipHeight() - startNode.Height) + 1)
-	if stopNode != nil && hs.Contains(stopNode) &&
-		stopNode.Height >= startNode.Height {
+	// Check if hashStop is lower than first valid height
```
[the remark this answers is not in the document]
jakubmkowalski: removed

transports/p2p/serverpeer.go
```diff
@@ -208,14 +208,14 @@ func (sp *serverPeer) OnGetHeaders(_ *peer.Peer, msg *wire.MsgGetHeaders) {
 	// over with the genesis block if unknown block locators are provided.
 	//
 	// This mirrors the behavior in the reference implementation.
-	headers := sp.server.syncManager.Services.Headers.LocateHeaders(msg.BlockLocatorHashes, &msg.HashStop)
 
-	// Send found headers to the requesting peer.
-	blockHeaders := make([]*wire.BlockHeader, len(headers))
-	for i := range headers {
-		blockHeaders[i] = &headers[i]
+	headers, err := sp.server.syncManager.Services.Headers.LocateHeaders(msg.BlockLocatorHashes, &msg.HashStop)
```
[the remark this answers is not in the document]
jakubmkowalski: changes reverted

service/header_service_test.go
jakubmkowalski: Tests splited into happy path and error path

service/header_service_test.go
```diff
@@ -348,6 +349,170 @@ func TestMerkleRootConfirmations(t *testing.T) {
 	}
 }
 
+func TestLocateHeadersGetHeadersHappyPath(t *testing.T) {
+	tData := setUpServices()
```
wregulski, line 353: regarding tests - please remember to run regression-tests before merging

dorzepowski (approved):

PR #261, opened 2024-06-28
ci(SPV-906): use split linter configuration in reusable workflow

dorzepowski:
# Pull Request Checklist


chris-4chain (approved):
👍

cmd/main.go
```diff
@@ -146,7 +141,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	var p2pServer P2PServer
+	var p2pServer p2pServer
```
ac4ch, line 144: well i think this is bad to have variable name and type same ?

config/config.go
```diff
@@ -61,16 +70,18 @@ type DbConfig struct {
 	// PreparedDbFilePath is the path to the prepared database file.
 	PreparedDbFilePath string `mapstructure:"prepared_db_file_path"`
 
-	Postgres PostgreSqlConfig `mapstructure:"postgres"`
+	Postgres PostgresqlConfig `mapstructure:"postgres"`
```
ac4ch, line 73: should not be `PostgreSQLConfig` ?

ac4ch: go is recommending to use  capital letters for acronyms

config/config.go
```diff
@@ -61,16 +70,18 @@ type DbConfig struct {
...
 	Sqlite   SqliteConfig     `mapstructure:"sqlite"`
 }
 
+// SqliteConfig represents a sqlite config.
 type SqliteConfig struct {
```
ac4ch, line 78: kind of same as in above SQL

config/defaults.go
```diff
@@ -91,8 +92,8 @@ func getMetricsDefaults() *MetricsConfig {
 	}
 }
 
-func getPostgresDefaults() PostgreSqlConfig {
-	return PostgreSqlConfig{
+func getPostgresDefaults() PostgresqlConfig {
+	return PostgresqlConfig{
```
ac4ch, line 96: same acronym with capital letters SQL

database/database.go
```diff
@@ -60,8 +59,8 @@ func newDbAdapter(cfg *config.DbConfig) (dbAdapter, error) {
 	switch cfg.Engine {
 	case config.DBSqlite:
 		return &sqLiteAdapter{}, nil
-	case config.DBPostgreSql:
-		return &postgreSqlAdapter{}, nil
+	case config.DBPostgresql:
+		return &postgresqlAdapter{}, nil
```
ac4ch, line 63: same acronym

database/export.go
```diff
@@ -106,7 +105,7 @@ func ExportHeaders(cfg *config.AppConfig, log *zerolog.Logger) error {
 }
 
 func queryDatabaseTable(db *sqlx.DB, log *zerolog.Logger) (*sqlx.Rows, error) {
-	rows, err := db.Queryx(selectHeadersSql)
+	rows, err := db.Queryx(selectHeadersSQL)
```
ac4ch, line 108: as example here u doing with capital so why in other places not ?

database/import.go
```diff
@@ -218,10 +218,8 @@ func validateHeightUniqueness(db *sqlx.DB) error {
 	_, err := db.Exec(fmt.Sprintf("CREATE UNIQUE INDEX %s ON headers (height)", tmpIndex))
 	if err != nil {
 		return errors.New("height values are not unique(they should be just after import)")
-	} else {
-		if _, err = db.Exec(fmt.Sprintf("DROP INDEX %s;", tmpIndex)); err != nil {
-			return fmt.Errorf("height values are unique buy droping temporary index %s failed", tmpIndex)
-		}
+	} else if _, err = db.Exec(fmt.Sprintf("DROP INDEX %s;", tmpIndex)); err != nil {
+		return fmt.Errorf("height values are unique buy droping temporary index %s failed", tmpIndex)
 	}
```
ac4ch, line 223: not sure but i dont like this block in my opinion this looks better 
```go
if err != nil {
    return errors.New("height values are not unique; expected uniqueness immediately after import")
}

dropIndexSQL := fmt.Sprintf("DROP INDEX %s;", tmpIndex)
if _, err = db.Exec(dropIndexSQL); err != nil {
    return fmt.Errorf("height values are unique, but dropping temporary index %s failed: %v", tmpIndex, err)
}
```

database/postgresql_adapter.go
```diff
@@ -12,24 +12,22 @@ import (
...
 	// use blank import to use file source driver with the migrate package.
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 	"github.com/jmoiron/sqlx"
-
 	"github.com/lib/pq"
+	"github.com/rs/zerolog"
 )
 
-type postgreSqlAdapter struct {
+type postgresqlAdapter struct {
```
ac4ch, line 23: name SQL

config/config.go
dorzepowski: image

dorzepowski: That's why I change it this way
But I see that they are not consistent about this name, sometimes they have Postgresql and sometimes it is PostgreSQL. If you don't mind, I would leave it as it is, because I don't want to go again and change every Posgresql into PostgreSQL

database/sql/headers.go
```diff
@@ -473,12 +472,9 @@ func (h *HeadersDb) getMerkleRootConfirmation(item domains.MerkleRootConfirmatio
...
 
-	if err != nil {
-		if errors.Is(err, sql.ErrNoRows) {
-		} else {
-			return nil, err
-		}
-	} else {
+	if err != nil && !errors.Is(err, sql.ErrNoRows) {
+		return nil, err
+	} else if err == nil {
```
ac4ch, line 477: same bad readability and finishing `if` block with `else if` not good idea
what about this 
```go
if err != nil {
    if !errors.Is(err, sql.ErrNoRows) {
        return nil, err
    }
}

// At this point, err is either nil or sql.ErrNoRows, handle the nil case 
if err == nil {
    confirmation.Hash = hash
}
return confirmation, nil  
```

database/export.go
dorzepowski: Answered about postgres

ac4ch (changes requested):
please corect SQL acronym  and avoid of finishing `if` block with `else if`

config/config.go
ac4ch: well, it not consistent and the name contain SQL up to u ... i posted what i think

database/sql/headers.go
dorzepowski: For me, your proposition has still readability issues. I will look for something better. But for me it's matter of taste, and probably we can discuss this for hours and won't came into a conclusion.

config/config.go
dorzepowski: Ok, I will change it, because the group behind it is `The PostgreSQL Global Development Group` so seems that SQL is more correct.

ac4ch (approved):

